PR #1, opened 2019-10-11
Add URL query string tools

roman-mazur:
AnalyticsUrlParam class encapsulates the logic of handling
passed URL parameters with anonymous device identifier.

Change-type: minor

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,5 +20,5 @@ const analytics = new AnalyticsUrlParams(mixpanel);
 
 analytics.consumeUrlParameters(window.location.search);
 
-const signupUrl = '/signup?' + analytics.deviceIdQuery();
+const signupUrl = '/signup?' + analytics.getDeviceIdsQueryString();
 ```
--- a/README.md
+++ b/README.md
@@ -1,2 +1,24 @@
-# analytics-client
-Client side analytics tools
+Analytics client
+================
+
+Client part of analytics services used at balena.
+
+## Installation
+
+```
+npm install --save analytics-client
+```
+
+## Usage
+
+URL query parameters tool.
+
+```typescript
+import { AnalyticsUrlParams } from 'analytics-client';
+
+const analytics = new AnalyticsUrlParams(mixpanel);
+
+analytics.consumeUrlParameters(window.location.search);
+
+const signupUrl = '/signup?' + analytics.deviceIdQuery();
+```
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
   },
   "dependencies": {
     "@analytics/cookie-utils": "^0.2.2",
-    "lodash": "^4.17.15",
     "typescript": "^3.6.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,0 +1,54 @@
+{
+  "name": "analytics-client",
+  "version": "0.0.1",
+  "description": "Convenient builders to compose analytics tools",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/balena-io/analytics-client"
+  },
+  "license": "Apache-2.0",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "test": "jest",
+    "prettify": "prettier --config ./node_modules/resin-lint/config/.prettierrc --write *.ts \"{src,test,typings}/**/*.ts\"",
+    "lint": "resin-lint --typescript src/ test/ && tsc --noEmit",
+    "build": "npm run test && tsc && webpack",
+    "prepublish": "npm run build"
+  },
+  "dependencies": {
+    "@analytics/cookie-utils": "^0.2.2",
+    "lodash": "^4.17.15",
+    "typescript": "^3.6.3"
+  },
+  "devDependencies": {
+    "@types/jest": "^24.0.18",
+    "@types/lodash": "^4.14.142",
+    "@types/mixpanel-browser": "^2.23.1",
+    "husky": "^3.0.9",
+    "jest": "^24.9.0",
+    "lint-staged": "^9.4.2",
+    "mixpanel-browser": "^2.29.1",
+    "resin-lint": "^3.1.0",
+    "ts-jest": "^24.1.0",
+    "ts-loader": "^6.2.0",
+    "webpack": "^4.41.1",
+    "webpack-cli": "^3.3.9"
+  },
+  "lint-staged": {
+    "*.ts": [
+      "prettier --config ./node_modules/resin-lint/config/.prettierrc --write",
+      "resin-lint --typescript --no-prettier",
+      "git add"
+    ],
+    "test/**/*.ts": [
+      "resin-lint --typescript --no-prettier --tests"
+    ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "pre-push": "npm run lint"
+    }
+  }
+}
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -3,7 +3,6 @@ import {
 	hasCookieSupport,
 	setCookie,
 } from '@analytics/cookie-utils';
-import union from 'lodash/union';
 import { Mixpanel } from 'mixpanel-browser';
 
 const URL_PARAM_DEVICE_ID = 'd_id';
@@ -17,7 +16,7 @@ const deviceIdSeparator = /\s*,\s*/;
  */
 export class AnalyticsUrlParams {
 	private readonly cookiesSupported = hasCookieSupport();
-	private deviceIds: string[];
+	private deviceIds: Set<string> = new Set();
 
 	constructor(private mixpanel?: Mixpanel) {
 		const storedValue = this.cookiesSupported
@@ -34,9 +33,9 @@ export class AnalyticsUrlParams {
 		if (currentDeviceId) {
 			list.push(currentDeviceId);
 		}
-		this.deviceIds = union(this.deviceIds, list);
+		this.deviceIds = new Set(list.concat(Array.from(this.deviceIds)));
 		if (this.cookiesSupported) {
-			setCookie(COOKIES_DEVICE_IDS, this.deviceIds.join(','));
+			setCookie(COOKIES_DEVICE_IDS, Array.from(this.deviceIds).join(','));
 		}
 		return list.length > 0 ? list[0] : null;
 	}
@@ -88,8 +87,13 @@ export class AnalyticsUrlParams {
 	 * @return all anonymous device IDs that can be passed to other sites
 	 */
 	allDeviceIds() {
-		const mixpanelId = this.mixpanel ? [this.mixpanel.get_distinct_id()] : [];
-		return union(this.deviceIds, mixpanelId);
+		const mixpanelId = this.mixpanel ? this.mixpanel.get_distinct_id() : null;
+		if (mixpanelId == null) {
+			return Array.from(this.deviceIds);
+		}
+		const res = new Set(this.deviceIds);
+		res.add(mixpanelId);
+		return Array.from(res);
 	}
 
 	/**
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -23,10 +23,7 @@ export class AnalyticsUrlParams {
 		const storedValue = this.cookiesSupported
 			? getCookie(COOKIES_DEVICE_IDS)
 			: null;
-		this.setDeviceIds(
-			storedValue,
-			this.mixpanel ? this.mixpanel.get_distinct_id() : null,
-		);
+		this.setDeviceIds(storedValue, null);
 	}
 
 	private setDeviceIds(
@@ -91,7 +88,8 @@ export class AnalyticsUrlParams {
 	 * @return all anonymous device IDs that can be passed to other sites
 	 */
 	allDeviceIds() {
-		return this.deviceIds;
+		const mixpanelId = this.mixpanel ? [this.mixpanel.get_distinct_id()] : [];
+		return union(this.deviceIds, mixpanelId);
 	}
 
 	/**
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -1,0 +1,107 @@
+import {
+	getCookie,
+	hasCookieSupport,
+	setCookie,
+} from '@analytics/cookie-utils';
+import union from 'lodash/union';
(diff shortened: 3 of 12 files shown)


nazrhom:
@balena-ci retest

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest please

webpack output is re-configured

roman-mazur:
Example of the compiled bundle served via unpkg:
https://unpkg.com/analytics-client@0.1.0-roman-url-params-431f6998b0134ff6209577529d1d3737fc8b1bbe/dist/bundle.js

webpack.config.js
```diff
@@ -0,0 +1,24 @@
+const path = require('path');
+
+module.exports = {
```
sradevski, line 3: Can we achieve the same only using the typescript compiler? Why do we need to use webpack?

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "lint": "resin-lint --typescript src/ test/ && tsc --noEmit",
+    "build": "npm run test && webpack",
+    "prepublish": "npm run build"
+  },
+  "dependencies": {
+    "@analytics/cookie-utils": "^0.2.2",
+    "lodash": "^4.17.15",
+    "typescript": "^3.6.3"
+  },
+  "devDependencies": {
```
sradevski, line 23: Can you add `lint-staged` and `husky` so we run linting and prettifying on commit?

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+	hasCookieSupport,
+} from '@analytics/cookie-utils';
+import { Mixpanel } from 'mixpanel-browser';
+
+const URL_PARAM_DEVICE_ID = 'd_id';
+const COOKIES_DEVICE_IDS = '__analytics_dids';
+
+const deviceIdSeparator = /\s*,\s*/;
+
+const cookiesSupported = hasCookieSupport();
```
sradevski: Maybe it's better to run this inside the constructor. I guess the environment won't change after loading the script, but it's better if the check is delayed only when the consumer decides they want to initialize the script.

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+	 *
+	 * @param queryString URL query string to process
+	 * @return a modified query string that hides processed/consumed parameters if the query string should be modified
+	 */
+	consumeUrlParameters(queryString: string): string | null {
+		const params = new URLSearchParams(queryString);
+
+		const passedDeviceId = params.get(URL_PARAM_DEVICE_ID);
+		if (passedDeviceId) {
+			let originalMpId: string | null = null;
```
sradevski: I would prefer if this is a full name `originalMixpanelId` as `Mp` can be confusing without thinking about it.

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+	 * Analyzes the query string and stores the anonymous device ID if it's there.
+	 * Returns null if the current URL query string should not be modified after consuming the data.
+	 *
+	 * @param queryString URL query string to process
+	 * @return a modified query string that hides processed/consumed parameters if the query string should be modified
+	 */
+	consumeUrlParameters(queryString: string): string | null {
+		const params = new URLSearchParams(queryString);
+
+		const passedDeviceId = params.get(URL_PARAM_DEVICE_ID);
```
sradevski, line 59: I would suggest parsing the parameter here from `string` to `string[]`, remove `takeFirstId` as it is a bit pointless at that point, push the original Mixpanel ID here, and pass only the parsed array to `storeInboundDeviceIds`. I feel like this separates the concerns of each function more clearly.

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * AnalyticsUrlParams helps with handling analytics-related URL parameters
+ * like anonymous device identifier.
+ */
+export class AnalyticsUrlParams {
+	private deviceIds: string | null;
+
+	constructor(private mixpanel?: Mixpanel) {}
+
+	private storeInboundDeviceIds(ids: string, currentDeviceId: string | null) {
```
sradevski:
```suggestion
	private setDeviceIds(ids: string, currentDeviceId: string | null) {
```

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+		}
+		const res = this.retrieveAndMerge(list);
+		this.deviceIds = res;
+		if (cookiesSupported) {
+			setCookie(COOKIES_DEVICE_IDS, res);
+		}
+		return res;
+	}
+
+	private retrieveAndMerge(input: string[]) {
```
sradevski: If you do what I suggested in the constructor, this won't be necessary (you can just use `this.deviceIds` directly). Then, the caller will turn it into a string if they wish to.

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+const cookiesSupported = hasCookieSupport();
+
+/**
+ * AnalyticsUrlParams helps with handling analytics-related URL parameters
+ * like anonymous device identifier.
+ */
+export class AnalyticsUrlParams {
+	private deviceIds: string | null;
+
+	constructor(private mixpanel?: Mixpanel) {}
```
sradevski: How about you check here if there are deviceIds in cookies, and set the class `deviceIds` from there? Like that you can only reference the class property, and only the cookies when you are setting.

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+			return params.toString();
+		} else {
+			return null;
+		}
+	}
+
+	/**
+	 * @return all anonymous device IDs that can be passed to other sites, separated with comma
+	 */
+	allDeviceIds(): string {
```
sradevski: I think it is better to return a `string[]` instead of the already concatenated string. You can have a separate method that will do the concatenation, otherwise the name is confusing (judging by the name, I would expect an array).

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+		if (this.deviceIds == null) {
+			const knownIds = this.mixpanel == null ? [] : [ this.mixpanel.get_distinct_id() ];
```
sradevski: Can we also do this in the constructor? If so, we can have `getDeviceIds` function that will simply be a getter for `this.deviceids`.

src/url-params.ts
```diff
@@ -0,0 +1,111 @@
...
+			this.deviceIds = this.retrieveAndMerge(knownIds);
+		}
+		return this.deviceIds!;
+	}
+
+	/**
+	 * @return part of a query parameter string that can be appended to URLs
+	 */
+	deviceIdQuery(): string {
```
sradevski:
```suggestion
	getDeviceIdsQueryString(): string {
```

sradevski (commented):
I think the url params class can be simplified, as currently there are a couple sources of truth (cookies, the local `deviceIds` and some functions append the mixpanel id to the result. If something is confusing, let me know and we can go through it together.

webpack.config.js
roman-mazur: We are using webpack to create a `bundle.js` (that contains all necessary pieces of this library) loadable by our blog and docs (via unpkg).

package.json
roman-mazur: On it...

webpack.config.js
sradevski: There is also `--outFile` option for the typescript compiler, but I just read there are some gotchas with it, so webpack might be the way to go.

src/url-params.ts
roman-mazur: Renamed.

package.json
roman-mazur: Added.

src/url-params.ts
roman-mazur: Moved to the constructor.

src/url-params.ts
roman-mazur: Renamed.

src/url-params.ts
roman-mazur: PTAL at the current implementation.

src/url-params.ts
roman-mazur: Renamed.

src/url-params.ts
roman-mazur: This function is removed now.

src/url-params.ts
roman-mazur: Check the current impl.

src/url-params.ts
roman-mazur: Yep, I agree.
this method now returns an array

src/url-params.ts
roman-mazur: It now returns the deviceId property value.

roman-mazur:
@sradevski your comments should be addressed now.

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest again, please
both TS and webpack outputs are now in the dist folder

roman-mazur:
@balena-ci retest please, again

README.md
```diff
@@ -1,2 +1,24 @@
...
+URL query parameters tool.
+
+```typescript
+import { AnalyticsUrlParams } from 'analytics-client';
+
+const analytics = new AnalyticsUrlParams(mixpanel);
+
+analytics.consumeUrlParameters(window.location.search);
+
+const signupUrl = '/signup?' + analytics.deviceIdQuery();
```
sradevski: Update this with the updated method name.

src/url-params.ts
```diff
@@ -0,0 +1,105 @@
...
+		} else {
+			return null;
+		}
+	}
+
+	/**
+	 * @return all anonymous device IDs that can be passed to other sites
+	 */
+	allDeviceIds() {
+		const mixpanelId = this.mixpanel ? [this.mixpanel.get_distinct_id()] : [];
```
sradevski: I think we can also do this check and assignment in the constructor, since that's where we get mixpanel passed. In that case we can have this method to be `getAllDeviceIds` and it will simply return `this.deviceIds`.

src/url-params.ts
```diff
@@ -0,0 +1,105 @@
...
+	 */
+	consumeUrlParameters(queryString: string): string | null {
+		const params = new URLSearchParams(queryString);
+
+		const passedDeviceId = params.get(URL_PARAM_DEVICE_ID);
+		if (passedDeviceId) {
+			const originalMixpanelId =
+				this.mixpanel != null ? this.mixpanel.get_distinct_id() : null;
+
+			const newCurrentDeviceId = this.setDeviceIds(
```
sradevski, line 64: Normally you wouldn't expect a setter to return anything, and in this case returning the first element is a bit confusing. How about 
```
this.setDeviceIds(passedDeviceIds);
const newCurrentDeviceId = this.deviceIds ? this.deviceIds[0] : null;
```

If you set `this.deviceIds` to the mixpanel distinct ID in the constructor, `setDeviceIds` will only require a single parameter, as `mixpanelId` will already by in `deviceIds`.

sradevski (commented):
A couple more comments to further simplify the code, the rest is 👍

src/url-params.ts
roman-mazur: See https://github.com/balena-io/analytics-client/pull/1/commits/bc5b491cad22dd7a8de7498a68b5772046bf4285

What you describe was my initial implementation, but then I changed it in the commit above making sure we don't expect mixpanel library to be initialized in the constructor.

src/url-params.ts
roman-mazur: Please note that this is not a public setter, but a private method and an implementation detail.

If we remove parsing of the input string, we'll need to parse it separately in the constructor (because cookie value is a string) and in `consumeUrlParameters`.
Since it does the parsing, it also returns the first device ID that can be used to configure mixpanel.

That said, we can rename it so it's not confused with a setter semantics. To something like `updateDeviceIds`.

sradevski: I think whether a function is public or private doesn't matter much in terms of functions being intuitive in their behavior. In this case, you will just read from `this.deviceIds` after calling `setDeviceIds` instead of expecting a return parameter, that's all.

src/url-params.ts
sradevski: If mixpanel is not passed in the constructor, there is no way to update it afterwards, as it is private. If a mixpanel client is indeed passed to the constructor, but it is not ready yet, then `this.mixpanel` will always return true, so this check will be wrong. Maybe I am missing something?

roman-mazur: A typical use of mixpanel:
```
import mixpanel from 'mixpanel-browser';

// AnalyticsUrlParams can be initialized at this moment of time.

mixpanel.init(token, options);

// Only now you can
mixpanel.get_distinct_id();
```

So we can do `const analyticsTool = new AnalyticsUrlParams(mixpanel)`, but `init ` is not yet called on the mixpanel object.

sradevski: But `mixpanel` is never falsy, so `this.mixpanel ? ... : ...` will always be true. I think we need to do something like `this.mixpanel.isInitialized` if there is such thing.

src/url-params.ts
roman-mazur: >I think whether a function is public or private doesn't matter much in terms of functions being intuitive in their behavior

My main point is not about public/private, but about the fact that it's not necessarily a classic setter. 

Also, although `this.deviceId` is an array, it's really a set of device IDs: when setDeviceIds is called, we merge the existing list with a new one, removing duplicates. The order of elements is not really guaranteed, hence, we should not assume `this.deviceIds[0]` is the first ID from the URL query.

README.md
roman-mazur: updated

src/url-params.ts
roman-mazur: The constructor is declared in a way it allows the mixpanel object to be skipped.
The logic handles this particular case (so it works when we do `new AnalyticsUrlParams()`).
The main point is that `get_distinct_id` call should be delayed/lazy, which is why `allDeviceIds` method is a bit more complicated.

Speaking about `isInitialized`, there is [no such method](https://github.com/mixpanel/mixpanel-js/issues/133) in mixpanel.

roman-mazur:
@balena-ci retest please

roman-mazur:
lodash dependency is removed now, `Set` is used to maintain device IDs.

@balena-ci retest, please

sradevski (approved):

PR #6, opened 2020-03-13
Replace cookies implementation

roman-mazur:
We need more functionality from the cookies package
(for, instance, listing all available cookies).

Change-type: minor


roman-mazur:
@balena-ci retest please

PR #8, opened 2020-03-17
Add experiments implementation

roman-mazur:
Change-type: minor

### DIFF
--- a/src/experiment.ts
+++ b/src/experiment.ts
@@ -1,0 +1,125 @@
+import { Identify } from 'amplitude-js';
+import { Client } from './client';
+
+const LOCAL_STORAGE_EXPERIMENTS_PREFIX = '__analytics_experiments_';
+
+/** Experiment interface allows getting a variation for a specified user. */
+export interface Experiment<Variation extends string> {
+	name: string;
+	engage(userId: string): Variation;
+}
+
+interface VariationData<Variation> {
+	variation: Variation;
+	targetPercent: number;
+}
+
+const checkPercent = (f: number) => {
+	if (isNaN(f) || f < 0 || f > 100) {
+		throw new Error(
+			'Variation target percent must be defined as a percent value between 0 and 100',
+		);
+	}
+};
+
+/** LocalExperiment allows describing an Experiment implemented with local storage. */
+export class LocalExperiment<Variation extends string>
+	implements Experiment<Variation> {
+	private readonly data: Array<VariationData<Variation>> = [];
+	private coveredPercent: number = 0;
+
+	constructor(
+		public readonly name: string,
+		private readonly analytics?: Client,
+	) {}
+
+	private checkDuplicates(variation: Variation) {
+		const present = this.data.find(d => d.variation === variation);
+		if (present != null) {
+			throw new Error(
+				`Variation [${present.variation} ${present.targetPercent}%] already exists in experiment ${this.name}.`,
+			);
+		}
+	}
+
+	private static dataString(data: Array<VariationData<any>>): string {
+		return data
+			.map(d => `variation ${d.variation}: ${d.targetPercent}%`)
+			.join(', ');
+	}
+
+	define(
+		variation: Variation,
+		targetPercent: number,
+	): LocalExperiment<Variation> {
+		checkPercent(targetPercent);
+		this.checkDuplicates(variation);
+		const varData = { variation, targetPercent };
+
+		this.coveredPercent += targetPercent;
+		if (this.coveredPercent > 100) {
+			const allData = LocalExperiment.dataString(this.data.concat(varData));
+			throw new Error(
+				`Incorrect target percent in experiment ${this.name}. Sum of fractions is greater than 100%: ${allData}`,
+			);
+		}
+
+		this.data.push(varData);
+		return this;
+	}
+
+	engage(deviceId: string): Variation {
+		if (this.data.length === 0) {
+			throw new Error(`Variations are not defined for experiment ${this.name}`);
+		}
+		if (this.coveredPercent < 100) {
+			throw new Error(
+				`Experiment ${
+					this.name
+				} is not fully defined. Current data: ${LocalExperiment.dataString(
+					this.data,
+				)}`,
+			);
+		}
+
+		if (window?.localStorage == null) {
+			// No storage support. Return a consistent result.
+			return this.data[0].variation;
+		}
+
+		const key = `${LOCAL_STORAGE_EXPERIMENTS_PREFIX}${this.name}_${deviceId}`;
+		const value = window.localStorage.getItem(key);
+		if (value != null) {
+			return value as Variation;
+		}
+
+		const dieRoll = Math.random() * 100;
+		let result: Variation | null = null;
+		let margin = 0;
+		for (const varData of this.data) {
+			margin += varData.targetPercent;
+			if (dieRoll < margin) {
+				result = varData.variation;
+				break;
+			}
+		}
+		if (result == null) {
+			throw new Error(
+				`Variations implementation problem: ${LocalExperiment.dataString(
+					this.data,
+				)}`,
+			);
+		}
+
+		window.localStorage.setItem(key, result);
+		if (this.analytics != null) {
+			this.analytics
+				.amplitude()
+				.identify(
+					new Identify().append('Experiments', `${this.name}_${result}`),
+				);
+		}
+
+		return result;
+	}
+}
--- a/test/experiment.test.ts
+++ b/test/experiment.test.ts
@@ -1,0 +1,84 @@
+import { createClient } from '../src/client';
+import { LocalExperiment } from '../src/experiment';
+
+describe('LocalExperiment', () => {
+	test('define bad fractions', () => {
+		expect.assertions(2);
+		try {
+			const exp = new LocalExperiment<'var1' | 'var2'>('test');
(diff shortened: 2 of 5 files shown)


roman-mazur:
@balena-ci retest please

test/experiment.test.ts
```diff
@@ -0,0 +1,84 @@
...
+		try {
+			exp.engage('test-user');
+		} catch (e) {
+			expect(e.message).toContain('fully');
+		}
+	});
+
+	describe('two variations experiment', () => {
+		const twoVariants = new LocalExperiment<'var1' | 'var2'>('test');
+		twoVariants.define('var1', 50).defineDefault('var2');
```
roman-mazur: @sradevski This is a proposal on how we can define the GSG modal experiment. It should be sufficient to run this experiment backed it with localStorage. And we can later hook Optimizely or our own backend implementation.

```
const gsgExp = new LocalExperiment<'ModalPresented'|'NoModal'>('GettingStartedGuide', analyticsClient);
const variation = gsgExp.engage(user.username);
if (variation === 'ModalPresented') {
  showModal();
}
```

test/experiment.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+		const exp = new LocalExperiment<'var1' | 'var2'>(
+			'amplitude-test',
+			analytics,
+		);
+		exp.define('var1', 30).define('var2', 70);
+
+		it('sets user property', () => {
+			exp.engage('test-user-1');
+			expect(identifyCallsCount).toStrictEqual(1);
```
roman-mazur, line 81: Analytics client will take care of setting the user property `Experiments=["${name}_${variation}"]`

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+		for (const varData of this.data) {
+			if (dieRoll < varData.margin) {
+				result = varData.variation;
+				break;
+			}
+		}
+		if (result == null) {
+			result = this.defaultVariation!;
+		}
+		window.localStorage.setItem(key, result);
```
sradevski, line 114: Is the analytics-client library only for the browser, or it also targets nodejs? If that's the case, we should check if window and localstorage exist so it doesn't throw. We should also prevent using experiments if it is a backend component, or accept an abstract storage parameter and use that.

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
+import { Identify } from 'amplitude-js';
+import { Client } from './client';
+
+const LOCAL_STORAGE_EXPERIMENTS_PREFIX = '__analytics_experiments_';
+
+/** Experiment interface allows getting a variation for a specified user. */
+export interface Experiment<Variation extends string> {
+	name: string;
+
```
sradevski: Let's remove the empty space

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+	private static checkPercent(f: number) {
+		if (isNaN(f) || f < 0 || f > 100) {
+			throw new Error(
+				'Variation fraction must be defines as a percent value between 0 and 100',
+			);
+		}
+	}
+
+	private checkDuplicates(variation: Variation) {
+		const present = this.data.find(d => d.variation === variation);
```
sradevski, line 37:
```suggestion
		const isPresent = this.data.some(d => d.variation === variation);
```

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+/** Experiment interface allows getting a variation for a specified user. */
+export interface Experiment<Variation extends string> {
+	name: string;
+
+	engage(userId: string): Variation;
+}
+
+interface VariationData<Variation> {
+	variation: Variation;
+	fraction: number;
```
sradevski: Let's rename this to something more descriptive, such as `targetPercentage`, as `fraction` doesn't imply the value is between 0 and 100, it can be anything.

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+
+	define(variation: Variation, fraction: number): LocalExperiment<Variation> {
+		LocalExperiment.checkPercent(fraction);
+		this.checkDuplicates(variation);
+
+		let previousMargin = 0;
+		if (this.data.length > 0) {
+			previousMargin = this.data[this.data.length - 1].margin;
+		}
+		const margin = previousMargin + fraction;
```
sradevski: The `margin` handling is quite strange, why do we need to keep it as part of each experiment variation? Let's make it a field of the `LocalExperiment` class that gets incremented, or even better, remove it completely. Also, why would we even want to support experiments that don't add up to 100%? What do we show to the rest if all experiments don't add up to 100%?

What might be better is to accept an array of `VariationData` in the `define` method, which will include the default variation as well, and check if the sum of those adds up to 100%. After all, `default` is not a special case, so there is no need to treat it as such.

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+			);
+		}
+
+		if (window.localStorage == null) {
+			// No storage support.
+			console.log(userId);
+			return this.defaultVariation || this.data[0].variation;
+		}
+
+		const key = `${LOCAL_STORAGE_EXPERIMENTS_PREFIX}${this.name}_${userId}`;
```
sradevski: I would probably not specify the user here, but the device (which means we don't need to add any suffix). I might have 2-3 accounts that I use on the same machine, and seeing different content based on the account I use might be strange.

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+			this.defaultVariation == null
+		) {
+			throw new Error(
+				`Experiments is not fully defined. Current data: ${LocalExperiment.dataString(
+					this.data,
+				)}`,
+			);
+		}
+
+		if (window.localStorage == null) {
```
sradevski: On non-browser environments `window` will be undefined, so 
```suggestion
		if (window?.localStorage == null) {
```

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+			return this.defaultVariation || this.data[0].variation;
+		}
+
+		const key = `${LOCAL_STORAGE_EXPERIMENTS_PREFIX}${this.name}_${userId}`;
+		const value = window.localStorage.getItem(key);
+		if (value != null) {
+			return value as Variation;
+		}
+
+		const dieRoll = Math.random() * 100;
```
sradevski, line 96: This is not really a die roll, since you get values between 0 and 100 :)

test/experiment.test.ts
sradevski: Yeah I think this is reasonable and simple enough to get us started quickly.

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+			throw new Error(
+				`Incorrect fraction in experiment ${this.name}. Sum of fractions is greater than 100%: ${allData}`,
+			);
+		}
+
+		this.data.push(varData);
+		return this;
+	}
+
+	defineDefault(variation: Variation): LocalExperiment<Variation> {
```
sradevski: Per my comment above, I don't think we need to treat `default` in a special way, which will simplify the code and how we reason about the library.

src/experiment.ts
```diff
@@ -0,0 +1,127 @@
...
+		const key = `${LOCAL_STORAGE_EXPERIMENTS_PREFIX}${this.name}_${userId}`;
+		const value = window.localStorage.getItem(key);
+		if (value != null) {
+			return value as Variation;
+		}
+
+		const dieRoll = Math.random() * 100;
+		let result: Variation | null = null;
+		for (const varData of this.data) {
+			if (dieRoll < varData.margin) {
```
sradevski: If we go with the approach I suggested, we can just find calculate the margin from the `data` itself, so we don't have to keep additional unnecessary state

src/experiment.ts
roman-mazur: It currently does not have a non-browser application. I would treat nodejs adaptation separately. It also requires more work on the amplitude client side.
I'm trying to keep the scope of this library small enough, for now, so we can address it later.

src/experiment.ts
roman-mazur: The search result is then used to construct an error message, so I would stick with the current variant.

src/experiment.ts
roman-mazur: Good point. I would still use an input deviceId as a suffix to allow getting different variations on the same device if needed (if we supply a different ID). But will rename the parameter and modify examples to provide the device ID, and not user ID.

roman-mazur:
@sradevski PTAL - I removed `default` and moved margin calculations to the variation detection code. Your comments should be addressed now.

roman-mazur:
@balena-ci retest please

src/experiment.ts
```diff
@@ -0,0 +1,125 @@
...
+		return data
+			.map(d => `variation ${d.variation}: ${d.targetPercent}%`)
+			.join(', ');
+	}
+
+	define(
+		variation: Variation,
+		targetPercent: number,
+	): LocalExperiment<Variation> {
+		LocalExperiment.checkPercent(targetPercent);
```
sradevski: Let's just take `checkPercent` outside the class

sradevski (commented):
Let's just take the function outside the class, LGTM otherwise.

src/experiment.ts
roman-mazur: Done.

roman-mazur:
@balena-ci retest please

sradevski (approved):

PR #13, opened 2020-03-31
Adjust compiler settings

roman-mazur:
Change-type: patch

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,11 +10,9 @@
     "preserveConstEnums": true,
     "removeComments": true,
     "resolveJsonModule": true,
-    "sourceMap": false,
+    "sourceMap": true,
     "strictNullChecks": true,
     "moduleResolution": "node",
-    "allowSyntheticDefaultImports": true,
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -10,11 +10,10 @@
     "preserveConstEnums": true,
     "removeComments": true,
     "resolveJsonModule": true,
-    "sourceMap": false,
+    "sourceMap": true,
     "strictNullChecks": true,
     "moduleResolution": "node",
-    "allowSyntheticDefaultImports": true,
-    "esModuleInterop": true,
+    "esModuleInterop": false,
```
thgreasi: Let's completely remove the line since that's the default (and I'm afraid when I see this in configs).

thgreasi (approved):
:+1: :+1: :+1: 

PS: I was literally about to push the same piece of code with the only difference being that I was using `import amplitude = require('amplitude-js');`.

roman-mazur:
@balena-ci retest please

PR #19, opened 2020-05-18
fix: pin mixpanel-browser version

JSReds:
pin dependency to specific version before breaking change

Change-type: patch
Signed-off-by: Andrea Rosci <andrear@balena.io>


roman-mazur (approved):

package.json
```diff
@@ -26,11 +26,11 @@
     "@types/jest": "^24.0.18",
     "@types/js-cookie": "^2.2.5",
     "@types/lodash": "^4.14.142",
-    "@types/mixpanel-browser": "^2.23.1",
+    "@types/mixpanel-browser": "2.23.1",
```
roman-mazur: We probably need to update the types version to get the tests working.

PR #20, opened 2020-05-26
Fix mixpanel dependency

roman-mazur:
We rely on it in runtime in order to detect a device ID.

Chang-type: patch


roman-mazur:
@balena-ci I self-certify!

ghost (approved):

PR #24, opened 2020-06-12
Call identify on re-engagement

iamsolankiamit:
Change-type: minor
Signed-off-by: Amit Solanki <amit@balena.io>

### DIFF
--- a/test/experiment.test.ts
+++ b/test/experiment.test.ts
@@ -80,6 +80,9 @@ describe('LocalExperiment', () => {
 		it('sets user property', () => {
 			exp.engage('test-device-1');
 			expect(identifyCallsCount).toStrictEqual(1);
+			exp.engage('test-device-1');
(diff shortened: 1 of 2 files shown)


test/experiment.test.ts
```diff
@@ -81,5 +81,11 @@ describe('LocalExperiment', () => {
 			exp.engage('test-device-1');
 			expect(identifyCallsCount).toStrictEqual(1);
 		});
+
+		it('sets user property on re-engage', () => {
```
roman-mazur: This test may be run in parallel with the previous one. Let's better move the second call directly to the original test:
```
exp.engage('test-device-1');
expect(identifyCallsCount).toStrictEqual(1);
exp.engage('test-device-1');
expect(identifyCallsCount).toStrictEqual(2);
```

roman-mazur (approved):
Thanks!

PR #25, opened 2020-06-17
Fix duplicate experiment values

roman-mazur:
With the existing implementation, we add a duplicate on every page reload.

Change-type: patch


roman-mazur:
@balena-ci I self-certify!

ghost (approved):

PR #29, opened 2020-07-29
Fix broken cookies

roman-mazur:
While working on integrating the latest client into our docs,
I noticed that Amplitude client is not able to create its cookie.
The issue is that in v5.11.0 they introduced the sameSiteCookie
property with default value None. However, it works only wih the
combination of setting Secure flag on a cookie in recent Chrome versions,
which is not done.

This change bumps the version, and fixes the behaviour setting
SameSite attribute to Lax. It matches the latest Amplitude lib
beehaviour (we don't update to the latest because of some breaking changes).

We also pin the client version to avoid such issues in the future.

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
     "prepublish": "npm run build"
   },
   "dependencies": {
-    "@types/amplitude-js": "^5.8.0",
-    "amplitude-js": "^5.10.0",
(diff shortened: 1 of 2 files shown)


roman-mazur:
I will do another bump for the balena-ui code.

sradevski (approved):

PR #30, opened 2020-11-20
Add documentation

gelbal:
Change-type: minor
Signed-off-by: Fırat Gelbal firat@balena.io


README.md
```diff
@@ -2,6 +2,8 @@ Analytics client
 ================
 
 Client part of analytics services used at balena.
+The client talks to [analytics-backend](https://github.com/balena-io/analytics-backend) which processes the incoming events before forwarding them to the relevant parties, referred here as *the final destinations*.
+Currently the only final destination is Amplitude. We have plans to support more final destinations going forward, like Sentry.
```
pranasziaukas: Ideally, as I see it, the final destinations should be handled in the `analytics-backend` as much as possible, and `analytics-client` should be just a friendly interface that doesn't have to worry about what happens next to ensure complete modularity.

In practice, it's probably next to impossible to avoid platform-specific (Mixpanel, Amplitude, Sentry) adaptations without reimplementing their SDKs - that's why we have these platforms mentioned in both `analytics-backend` and `analytics-client`.

README.md
```diff
@@ -60,3 +62,44 @@ Then use the `analyticsClient` variable:
...
+
+## Identifying users
+
+By default the client tracks `deviceId` to identify the current device of the end user. `deviceId` is a UUID generated automatically by the client. Use `setUserId` method to identify authenticated user. At least one of these identifiers is required to successfully track an event.
+
+In most cases the device refers to the browser of the user (or Electron application for the case of Etcher). By persisting the `deviceId` in the local storage, we are able to identify the events coming from the same user in order to create a timeline of component usage.
+
+It is possible to generate a new device identifier and also override an existing device identifier. We use these methods especially when we need to link multiple device identifiers that we know belonging to the same user.
```
pranasziaukas: Why is this needed? My impression was that different devices can belong to the same user - thus device ID and user ID are accounted for. Or is it actually the same thing? :)

README.md
```diff
@@ -60,3 +62,44 @@ Then use the `analyticsClient` variable:
...
+For instance, a balenaCloud user might login to the dashboard from multiple browsers and also use balenaCLI. We would like to create a coherent timeline that captures all their activity and display it under the same user in order to make a sound analysis of their balena platform usage. As long as all the components `setUserId` as soon as the authentication happens, we are able to link all the events with separate `deviceId` values that came *even before* the authentication on the backend side.
```
pranasziaukas, line 131: How does it synergize with the merging abilities of Amplitude? Or are we reimplementing the same thing?

pranasziaukas (commented):
Thank you for the documentation Firat, it is also very helpful for my understanding, even though I've seen pieces of conversations on FD.

Where should we seek for a quick unbiased review? Maybe @Page- could take a look?

README.md
pranasziaukas: Looking into the code more, it's probably about updating the collection of device IDs rather than overriding a single ID, is that right?

README.md
gelbal: Agreed Pranas. Is there anything you suggest changing in this paragraph?
I like your sentence with modularity. We could paraphrase and add it here.

I'd like to mention Amplitude and Sentry to put concrete use case in the beginning of readme. So the component maintainer immediately gets what it means to integrate analytics-client. For sure, we should detail the specific of these final destinations on the analytics-backend docs.

README.md
gelbal: Yes, you captured the common cases. The next paragraphs also expand on these.

There are some more to it: as we transition some components from Mixpanel to Amplitude, we prefer to re-use the Mixpanel identifiers as Amplitude identifiers (not to restart the events timeline of existing user). Also in the case of unauthenticated usage as in Etcher, we rely on `deviceId` to track the same user. So a component maintainer might need to use these methods mentioned to achieve this.

Shall we add anything extra to this paragraph?

README.md
gelbal: We are utilizing Amplitude SDK's methods to achieve these actually. The main logic for merging sits in url-params.ts file. There we parse the URL parameters to link multiple device ids to the same user id.

gelbal:
Thanks for the quick turnaround and questions Pranas.

gelbal:
Hey @Page- @sradevski @iamsolankiamit, it'd be great to get a quick review on the docs to see if these make sense. We are going to promote analytics-client and it helps to have a starting docs.

README.md
pranasziaukas: > we prefer to re-use the Mixpanel identifiers as Amplitude identifiers (not to restart the events timeline of existing user)

This is perfectly understandable, but does it involve overwriting the *device* ID specifically?

README.md
pranasziaukas: Maybe something along the following lines
```suggestion
The analytics client is a component that integrates into the [analytics-backend](https://github.com/balena-io/analytics-backend). Together they're responsible for processing the analytics tracking events before forwarding them to the relevant parties, referred to here as *the final destinations*.
Currently the only final destination is Amplitude. We have plans to support more final destinations going forward, like Sentry.
```

I don't think we should expand too much on the `analytics-backend` for the exact reasons of modularity and it being an independent component/repo :)

README.md
```diff
@@ -60,3 +62,44 @@ Then use the `analyticsClient` variable:
 ```js
 const urlParamsHandler = new analyticsClient.AnalyticsUrlParams();
 ```
+
+## Configuration
+
+We use environment configuration on the [analytics-backend](https://github.com/balena-io/analytics-backend) to mask the actual API tokens of the final destination services of these events.
```
sradevski, line 101: This section is quite verbose TBH and it includes a lot of internals the user doesn't have to worry about. How about just writing down a configuration example code, and then add comments per configuration entry (eg. `projectName - a unique name for your project. Reach out to the data team when in doubt about the name you should use`)

README.md
```diff
@@ -60,3 +62,44 @@ Then use the `analyticsClient` variable:
...
+
+Additionally, identify your component using the parameter `componentName`. It's best to keep the value short yet descriptive.
+
+Please ask Operation folks for `#access` to the relevant project in your final destination so you can validate that the tracked events show as expected.
+
+Finally, there is the optional configuration parameter `componentVersion` to track the version of a component sending each event. This is useful to track changes across new component versions so we strongly recommend tracking it.
+
+## Web tracker
+
+[`WebTracker`](src/web.ts) exposes an events tracking interface suitable for web apps. It provides a simple `Page View` event tracking with page loading metrics via `trackPageView`.
```
sradevski, line 118: Again, an example on how to use it with a short explanation would be nice here. In general, `examples >>> text` 😄

sradevski (commented):
Great job writing down some docs! In general examples are much more useful and easier to understand than a long explanation, and only use textual explanation when the code is not self-explanatory. Basically, we should write documentation the same way one would write code and comments. 👍

README.md
gelbal: Sounds good. Thanks Pranas.

README.md
gelbal: Thanks for the thoughts @sradevski. You raise a very good point. I'm adding couple of more examples to the top of the readme to illustrate the common use cases. Then is there any harm to keep such verbose explanations towards the bottom of the docs?

So the option is still there for anyone who wants to understand the internals.

README.md
gelbal: > does it involve overwriting the device ID specifically?

The answer depends on the setup. 

Now I see how this sentence is confusing. I'm pushing a reworked version Pranas, let me know if it's still confusing.

README.md
sradevski: I personally don't think it is needed, at least the internal implementation parts, since every single character or word we write is something that needs to be maintained and can get outdated. The less we write, the better (of course, without losing information)

gelbal: Good point on the maintenance part Stevche. 

I prefer to keep these explanations here until we find a better place. In a sense, these are the answers to FAQ raised in flowdock threads. If similar questions keep popping up, I'll take it as a strong signal that indeed no one reads these explanations here : )  
Also, going forward, it might make sense to keep such explanations in the repo but move them to a new section for the maintainers.

README.md
```diff
@@ -11,6 +13,46 @@ npm install --save analytics-client
...
+Tracking a page view. The following reports an event named `[EFP] Page View` with default properties.
+
+```typescript
+import { createClient, WebTracker, createWebTracker } from 'analytics-client';
+
+const client = createClient({
+    endpoint: 'data.balena-cloud.com', // use 'data.balena-staging.com' for testing the integration
+    projectName: 'balena-project', // unique identifier that analytics-backend expects
+    componentName: 'etcher-featured-project', // short unique name of component
+    componentVersion: '0.52', // automated version reporting of component is beneficial
```
Page-, line 25: What does it mean by automated version reporting? I assume you mean something like `require('./package.json').version` but that's really not conveyed by this comment

README.md
```diff
@@ -11,6 +13,46 @@ npm install --save analytics-client
 
 ## Usage
 
+Tracking a page view. The following reports an event named `[EFP] Page View` with default properties.
```
Page-, line 16: Where does the `EFP` bit come from? I can't see it below at all so I assume it's using some external context?

README.md
```diff
@@ -11,6 +13,46 @@ npm install --save analytics-client
...
+
+Here is an example how a product like CLI would identify a user and report a login event (`[CLI] Login`) with certain event properties related to the login action.
+
+```typescript
+import { createClient } from 'analytics-client';
+
+const client = createClient({
+    endpoint: 'data.balena-cloud.com', // use 'data.balena-staging.com' for testing the integration
+    projectName: 'balena-project', // unique identifier that analytics-backend expects
+    componentName: 'CLI', // short unique name of component
```
Page-, line 41: I'd probably avoid this an example tbh, `CLI` isn't what I'd normally think of as particularly unique (and if we were to do it today I'm sure we'd use eg `balena-cli`

README.md
pranasziaukas: [#L29](https://github.com/balena-io-modules/analytics-client/blame/575e4cf7091a190f8e72eab5ec70238de31d5d2d/README.md#L29) sets the prefix `createWebTracker(client, 'EFP')`.

README.md
pranasziaukas: A good observation. I believe we're mostly interested in referencing the state/version of the tracking implementation, and that would come with the component version by default.

README.md
```diff
@@ -11,6 +13,46 @@ npm install --save analytics-client
 
 ## Usage
 
+Tracking a page view. The following reports an event named `[EFP] Page View` with default properties.
+
+```typescript
+import { createClient, WebTracker, createWebTracker } from 'analytics-client';
```
pranasziaukas, line 19: `WebTracker ` is referenced but unused, isn't it?

README.md
```diff
@@ -60,3 +95,44 @@ Then use the `analyticsClient` variable:
...
+
+Please ask Operation folks for `#access` to the relevant project in your final destination so you can validate that the tracked events show as expected.
+
+Finally, there is the optional configuration parameter `componentVersion` to track the version of a component sending each event. This is useful to track changes across new component versions so we strongly recommend tracking it.
+
+## Web tracker
+
+[`WebTracker`](src/web.ts) exposes an events tracking interface suitable for web apps. It provides a simple `Page View` event tracking with page loading metrics via `trackPageView`.
+
+When initializing `createWebTracker`, set `prefix` as a short string that describes your project. All events tracked by `WebTracker` will have this `prefix` in their name.
```
pranasziaukas, line 120: I guess we wouldn't need a case with no prefix, right?

README.md
```diff
@@ -60,3 +95,44 @@ Then use the `analyticsClient` variable:
...
+
+[`AnalyticsUrlParams`](src/url-params.ts) helps with linking anonymous `deviceId` values that each component sends individually. For instance we use this method to link Etcher users clicking on EFP link to land on our blog website and eventually signup to balenaCloud. When the user clicks on the blog link in EFP, we append the `deviceId` generated by EFP component to the URL. Then the blog parses the URL to match EFP `deviceId` with blog `deviceId`. In a similar fashion, the signup links on the blog website also includes EFP `deviceId` as well as blog `deviceId`. When the user lands on balenaCloud dashboard and authenticates, we link all the user activity and list it under the same timeline. We need to make these jumps as there are 3 components tracking events and there are multiple domains involved. The benefit is we are able to look into a user's EFP and blog activity before signing up to balenaCloud.
```
pranasziaukas, line 133: As I understand, it means pretty tight integration with specific components appending extra information to the URL, and then only specific components at particular endpoints are expecting and parsing that extra information.
Is that an accurate description of our use case?

pranasziaukas (commented):
I quite like the current separation of "getting started" examples followed by more verbose descriptions and explanations, I think it does justice to the immediate needs.

README.md
pranasziaukas: BTW, maybe "blog `deviceId`" could be easier understood as "browser `deviceId`", no?

README.md
pranasziaukas: As discovered at the end of [this](https://www.flowdock.com/app/rulemotion/i-cli/threads/77cAobr10SuHLdgJPHntqRIMrn7) thread, CLI is currently not compatible with the `analytics-client` as it lacks support for the node environment.

Can we rename CLI to something else and recycle the example?

pranasziaukas:
@gelbal if you could share your opinion on the outstanding remarks, I think then we can go ahead and merg. Probably won't get any more significant feedback during this iteration.

sradevski (approved):

README.md
gelbal: Indeed right now only web tracker supports event name prefix. It's best to allow any tracker to configure such prefix. I created an issue about this: https://github.com/balena-io-modules/analytics-client/issues/36

README.md
gelbal: Indeed. I'll drop it from here as it creates confusion.

README.md
gelbal: Right now I guess we use prefix on all web trackers but it's good to keep the prefix optional. On a non web tracker, we already have a case with no prefix: Etcher events. Since Etcher is the only component that sends events to the Etcher project in Amplitude, there is no reason to have a prefix.

README.md
gelbal: Indeed it's a pretty tight integration Pranas. As we discussed last week with Stevche as well, it makes sense to improve the current state of implementation to make the whole device identifier linking flow more convenient for component maintainers.

README.md
gelbal: Good point. I'll rewrite the example here with `balena-cli`. I guess the repository name of a component is a good candidate to be `componentName`.

README.md
gelbal: > `require('./package.json').version`

Oh this is brilliant @Page-. I didn't know one can get the version like this. I'll update the example with it.

PR #39, opened 2021-02-03
Implement session ID parsing and logic

pranasziaukas:
My first real attempt to merge a TypeScript code change into a live production component.
Although we already have `deviceId` logic running along similar lines, there are some subtle differences (different type, a single value) that I tried to address and cover with unit tests.
Regardless, *please* take a close look to make sure that we're not breaking everything :)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -27,6 +27,9 @@ export interface Client {
 	/** Return the ID used to identify the current device. */
 	deviceId(): string;
 
+	/** Return the ID used to identify the current session. */
+	sessionId(): number;
+
 	/** Generate a new device identifier used for reporting. */
 	regenerateDeviceId(): void;
 
@@ -38,6 +41,9 @@ export interface Client {
 
 	/** Set current device ID. */
 	setDeviceId(deviceId: string): void;
+
+	/** Set current session ID. */
+	setSessionId(sessionId: number): void;
 
 	/** Set current user ID. */
 	setUserId(userId: string): void;
@@ -134,8 +140,16 @@ class DefaultClient implements Client {
 		return this.amplitudeInstance.options.deviceId!!;
 	}
 
+	sessionId(): number {
+		return this.amplitudeInstance.getSessionId();
+	}
+
 	setDeviceId(deviceId: string) {
 		this.amplitudeInstance.setDeviceId(deviceId);
+	}
+
+	setSessionId(sessionId: number) {
+		this.amplitudeInstance.setSessionId(sessionId);
 	}
 
 	regenerateDeviceId() {
@@ -213,6 +227,10 @@ class NoopClient implements Client {
 		return '';
 	}
 
+	sessionId() {
+		return -1;
+	}
+
 	linkDevices() {
 		/* nothing */
 	}
@@ -220,6 +238,9 @@ class NoopClient implements Client {
 		/* nothing */
 	}
 	setDeviceId(): void {
+		/* nothing */
+	}
+	setSessionId(): void {
 		/* nothing */
 	}
 	setUserId(): void {
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -2,7 +2,6 @@ import * as Cookies from 'js-cookie';
 import { Client } from './client';
 import {
 	COOKIES_DEVICE_IDS,
-	COOKIES_SESSION_ID,
 	COOKIES_TTL_DAYS,
 	URL_PARAM_DEVICE_ID,
 	URL_PARAM_OPT_OUT_REQUEST,
@@ -24,9 +23,6 @@ export class AnalyticsUrlParams {
 	constructor(private client?: Client) {
 		const storedDeviceIdValue = Cookies.get(COOKIES_DEVICE_IDS);
 		this.setDeviceIds(storedDeviceIdValue, null);
-
-		const storedSessionIdValue = Cookies.get(COOKIES_SESSION_ID);
-		this.setSessionId(storedSessionIdValue);
 	}
 
 	private setDeviceIds(
@@ -52,17 +48,12 @@ export class AnalyticsUrlParams {
 	private setSessionId(storedValue: string | null | undefined) {
 		if (storedValue) {
 			this.sessionId = Number(storedValue);
-			Cookies.set(COOKIES_SESSION_ID, storedValue, {
-				expires: COOKIES_TTL_DAYS,
-				path: '/',
-			});
 		}
 		return this.sessionId;
 	}
 
 	clearCookies() {
 		Cookies.remove(COOKIES_DEVICE_IDS);
-		Cookies.remove(COOKIES_SESSION_ID);
 	}
 
 	/**
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -2,9 +2,11 @@ import * as Cookies from 'js-cookie';
 import { Client } from './client';
 import {
 	COOKIES_DEVICE_IDS,
+	COOKIES_SESSION_ID,
 	COOKIES_TTL_DAYS,
 	URL_PARAM_DEVICE_ID,
 	URL_PARAM_OPT_OUT_REQUEST,
+	URL_PARAM_SESSION_ID,
 } from './config';
 
 const deviceIdSeparator = /\s*,\s*/;
@@ -16,11 +18,15 @@ const deviceIdSeparator = /\s*,\s*/;
 export class AnalyticsUrlParams {
 	private deviceIds: Set<string> = new Set();
 	private passedDeviceId: string | undefined;
+	private sessionId: number | undefined;
 	private optOutRequested: boolean = false;
 
 	constructor(private client?: Client) {
-		const storedValue = Cookies.get(COOKIES_DEVICE_IDS);
-		this.setDeviceIds(storedValue, null);
+		const storedDeviceIdValue = Cookies.get(COOKIES_DEVICE_IDS);
+		this.setDeviceIds(storedDeviceIdValue, null);
+
+		const storedSessionIdValue = Cookies.get(COOKIES_SESSION_ID);
+		this.setSessionId(storedSessionIdValue);
 	}
 
 	private setDeviceIds(
@@ -43,12 +49,24 @@ export class AnalyticsUrlParams {
 		return list.length > 0 ? list[0] : null;
 	}
 
+	private setSessionId(storedValue: string | null | undefined) {
+		if (storedValue) {
+			this.sessionId = Number(storedValue);
+			Cookies.set(COOKIES_SESSION_ID, storedValue, {
+				expires: COOKIES_TTL_DAYS,
+				path: '/',
+			});
+		}
+		return this.sessionId;
+	}
+
 	clearCookies() {
 		Cookies.remove(COOKIES_DEVICE_IDS);
+		Cookies.remove(COOKIES_SESSION_ID);
 	}
 
 	/**
-	 * Analyzes the query string and stores the anonymous device ID if it's there.
+	 * Analyzes the query string and stores the anonymous device or session ID if it's there.
 	 * Returns null if the current URL query string should not be modified after consuming the data.
 	 *
 	 * @param queryString URL query string to process
@@ -73,10 +91,21 @@ export class AnalyticsUrlParams {
 			}
 
 			params.delete(URL_PARAM_DEVICE_ID);
-			return params.toString();
-		} else {
-			return null;
 		}
+
+		const passedSessionId = params.get(URL_PARAM_SESSION_ID);
+		if (passedSessionId) {
+			const newCurrentSessionId = this.setSessionId(passedSessionId);
+
+			if (this.client != null && newCurrentSessionId != null) {
+				this.client.setSessionId(newCurrentSessionId);
+			}
+
+			params.delete(URL_PARAM_SESSION_ID);
+		}
+
+		const remainingQueryString = params.toString();
+		return queryString === remainingQueryString ? null : remainingQueryString;
 	}
 
 	setClient(client: Client) {
@@ -87,6 +116,7 @@ export class AnalyticsUrlParams {
 		if (newDeviceId != null) {
 			client.setDeviceId(newDeviceId);
 		}
+		this.sessionId = client.sessionId();
 		this.client = client;
 	}
 
@@ -111,7 +141,14 @@ export class AnalyticsUrlParams {
 	}
 
 	/**
-	 * @return part of a query parameter string that can be appended to URLs
+	 * @return session ID that can be passed to other sites
+	 */
+	getSessionId() {
+		return this.client ? this.client.sessionId() : this.sessionId;
+	}
+
+	/**
+	 * @return part of the device ID query parameter string that can be appended to URLs
 	 */
 	getDeviceIdsQueryString(): string {
 		const ids = this.allDeviceIds();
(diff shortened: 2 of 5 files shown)


src/client.ts
```diff
@@ -206,6 +220,10 @@ class NoopClient implements Client {
 		return '';
 	}
 
+	sessionId() {
+		return 0;
```
gelbal: I suggest setting this value to -1 instead as that's the default value of Amplitude. [Relevant docs](https://help.amplitude.com/hc/en-us/articles/115002323627-Tracking-Sessions#h_e9e9580e-5748-49ed-8ec4-8583c700ddd7).

src/url-params.ts
```diff
@@ -99,7 +129,14 @@ export class AnalyticsUrlParams {
 	}
 
 	/**
-	 * @return part of a query parameter string that can be appended to URLs
+	 * @return session ID that can be passed to other sites
+	 */
+	getSessionId() {
+		return this.sessionId;
```
gelbal:
```suggestion
		return this.client ? this.client.sessionId() : this.sessionId;
```

It's more reliable to depend on Amplitude as the single source of truth. Besides, `this.sessionId` would be undefined / null if there is no incoming `sessionId`.

src/url-params.ts
```diff
@@ -38,12 +44,24 @@ export class AnalyticsUrlParams {
 		return list.length > 0 ? list[0] : null;
 	}
 
+	private setSessionId(storedValue: string | null | undefined) {
+		if (storedValue) {
```
gelbal, line 49: I have a discussion point: Shall we follow Amplitude's logic and define timeout windows? 
[Relevant Amplitude docs](https://help.amplitude.com/hc/en-us/articles/115002323627-Tracking-Sessions#h_c6714eea-04a4-48ce-b2e0-e0379213ad95)

When the incoming `sessionId` is from 1 hour ago or longer, do we still want to apply it as the session of the new events?

gelbal: I suggest applying the incoming `sessionId` value only if it's in an acceptable time window, like 30 minutes or perhaps 1 hour.

src/url-params.ts
```diff
@@ -109,6 +146,17 @@ export class AnalyticsUrlParams {
 		return `${URL_PARAM_DEVICE_ID}=${encodeURIComponent(ids.join(','))}`;
 	}
 
+	/**
+	 * @return part of the session ID query parameter string that can be appended to URLs
+	 */
+	getSessionIdQueryString(): string {
+		const id = this.sessionId;
```
gelbal: Likewise, we need to update here to return `this.client.sessionId()` if it exists.

src/url-params.ts
```diff
@@ -109,6 +146,17 @@ export class AnalyticsUrlParams {
 		return `${URL_PARAM_DEVICE_ID}=${encodeURIComponent(ids.join(','))}`;
 	}
 
+	/**
+	 * @return part of the session ID query parameter string that can be appended to URLs
+	 */
+	getSessionIdQueryString(): string {
```
gelbal, line 155: What do you think defining an additional method here to consolidate the results of `getDeviceIdsQueryString` and `getSessionIdQueryString`?
Mainly for convenience reasons. So the component maintainer could retrieve everything in one go to append to the URL.

gelbal (commented):
Nice work Pranas. Great to see the unit tests. I kindly ask for couple of changes and a discussion around long running sessions.

src/url-params.ts
pranasziaukas: Sounds good.

src/client.ts
sradevski: How about returning `null` instead? I think it's more accurate to say that there is no session id, rather than returning an arbitrary number

src/url-params.ts
sradevski: It should be 30minutes ~ 1hour without any activity, if the users spends 4 hours on the dashboard by actively doing something, then that should be the same session.

src/client.ts
gelbal: Sure `null` works as well.

src/url-params.ts
gelbal: @sradevski right. Then we might need to record and check the timestamp of the navigation moment. `sessionId` is essentially the epoch timestamp of the first event in a session. All the following events in the same session use the same timestamp. So we could decide not to modify `sessionId` when a user was inactive for X minutes after navigating to the next component. 

I'm thinking particularly the case of users saving / sharing URLs with `sessionId` param. Though I'm not sure how large of an impact these will have, it'd be nice not to pollute our session data.

src/url-params.ts
pranasziaukas: I am adding it as an additional method `getQueryString`.
As I understand, there might be legitimate cases where we only pass device or session IDs so leaving those methods as they are seems wise (besides the obvious backwards-compatibility).

src/client.ts
pranasziaukas: Does it matter though?
Note that this is part of the `NoopClient` implementation, which is some kind of an empty mock, not even sure where it's used (maybe for prototyping?).

I've changed the return value to `-1` as per documentation. `null` would violate the expected `number` type - I assume `deviceId()` returns an empty string instead of `null` for similar reasons.

src/url-params.ts
pranasziaukas: I am familiar with the technique of splitting sessions after a certain amount of inactivity and used 30 min timeouts in previous projects.

If done by ourselves, it certainly means keeping track of the *latest* activity timestamp to decide if the subsequent event qualifies for the same session.
But quickly looking at the docs ([this](https://help.amplitude.com/hc/en-us/articles/115002323627-Tracking-Sessions#h_7a795c15-e644-4cec-9f99-54dc24972f27) section specifically) my understanding is that this should be handled by Amplitude automatically. If that's the case, I'd be hesitant to re-implement the same logic (a.k.a. reinvent the wheel).

Let me know if I'm missing any crucial aspects here 🤔

pranasziaukas:
@sradevski do you happen to know what's going on with the CI?

```
node_modules/@types/lodash/common/common.d.ts:234:132 - error TS2344: Type 'K' does not satisfy the constraint 'string'.
  Type 'string | number | symbol' is not assignable to type 'string'.
    Type 'number' is not assignable to type 'string'.

234     type MemoVoidDictionaryIterator<T, K extends string | number | symbol, TResult> = (acc: TResult, curr: T, key: K, dict: Record<K, T>) => void;
```

it looks like the linter is unhappy with `lodash`, which is weird.

src/url-params.ts
gelbal: Sounds like it's best to test the default behaviour here before we start adding extra logic. I don't know if / how Amplitude would behave when we apply a custom `sessionId`.

pranasziaukas: Good idea!

gelbal: @pranasziaukas I'm running both `analytics-client` and `analytics-backend` locally to observe Amplitude's behaviour around custom `sessionId`. Triggering from `/example/main.js` and using the default values, I could see my events collected under `Stg - test` project (specifically [this user](https://analytics.amplitude.com/balena/project/252097/search/amplitude_id%3D148200232010)). First of all, I suggest you to run the full stack locally yourself as well. I found it quite helpful. Moving on this `sessionId` discussion: seems like Amplitude happily merges all the events sent under the same `sessionId`. There is no extra logic based on timeout windows. 

[The top events](https://analytics.amplitude.com/balena/project/252097/search/amplitude_id%3D148200232010?sessionHandle=XDtXqKI_XeCIie6_CKBbBRK_9jB_0000-0000-0000&eventId=6d5fb40a-6a1d-11eb-8433-0a0cabafa3f1) on February 8 are merged to the session from March 18 (`session id = 1584530367112`). Next using `session id = 1612345671145`, I merged [today's events](https://analytics.amplitude.com/balena/project/252097/search/amplitude_id%3D148200232010?sessionHandle=XdnSn3p_XdnSn36_DXEG0b2_9jB_7db36c83-5eb7-4f56-a377-4d2a0923fac4R&eventId=20a0da18-6a1d-11eb-96b7-063a4d28002b) with the events from last week. 

So either on `analytics-client` or on `analytics-backend` it might help to introduce some extra logic ourselves to prevent such "session pollution". Would you agree?

sradevski:
@pranasziaukas try adding `"keyofStringsOnly": false,` to the tsconfig.json compiler options

pranasziaukas:
> @pranasziaukas try adding `"keyofStringsOnly": false,` to the tsconfig.json compiler options

Looks better already, thanks Stevche!

src/url-params.ts
pranasziaukas: Thanks for the instructions Firat, I'll give it a go tomorrow.

> So either on `analytics-client` or on `analytics-backend` it might help to introduce some extra logic ourselves to prevent such "session pollution". Would you agree?

I certainly agree that a sane session splitting between prolonged times of inactivity is very much useful and needed. It's just a bit disappointing if Amplitude's client cannot initiate a new session automatically. In that case, I'd at least like to better understand where this limitation is coming from as the documentation (in your first post above) describes default timeout windows of 30 mins, so we were not just speculating wildly.

pranasziaukas: @gelbal could you elaborate (even more :)) on running the test flow locally?

I ran into issues while starting the `analytics-backend` https://github.com/balena-io/analytics-backend/issues/26
Furthermore, my impression was that `/example/index.html` could be opened, which in turn loads the `main.js`. Do you do it differently?

gelbal: @pranasziaukas right. I see the same redis errors on my end. I ignored them as they were not relevant to our exercise here. `analytics-backend` uses redis to count and throttle the supervisor events that are triggered in small window of time. Good idea to open that issue so we document the devenv setup clearly. 

Anyhow, indeed I was editing `main.js` to test stuff.

gelbal: Thinking about this further: Shall we use the expiry date of cookies to achieve what we want? 
We could set a hard upper limit based on `sessionId` value and `currentTime`. The downsides of this is such hard limit surely impacts the analysis. Though this might be preferred actually.

pranasziaukas: @gelbal, do you mean setting the hard limit once and thus limiting how long a single session could be? I don't say it would be wrong but surely unusual as typically sessions last for as long as there's a continuous activity.

Or do you mean updating the cookie's TTL every time we fire an event? I have no idea if that would be a good practice, to be honest.

Both add to my hesitation to implement the full session logic ourselves. But maybe I just have to let the idea brew in my head :)

gelbal: > Or do you mean updating the cookie's TTL every time we fire an event?

Indeed, this is what I was thinking Pranas. My reasoning is not to have extra logic in the `analytics-client` code but instead utilize the Cookies expiry date to decide until when a passed session ID is valid. 

I'm curious to hear how you'd like to implement this. If it helps, we could raise it to Kimmo to see if he knows a common / best practice here.

pranasziaukas: In the ideal world, I'd like Amplitude's client to manage as much of this as possible i.e. to generate a new session ID after a certain period of inactivity for us.

An illustration:
- Events are being sent under the session X.
- A period of inactivity takes place.
- Events are being sent under the session Y (updated courtesy of Amplitude).
- At any point, we get the relevant session ID because `AnalyticsUrlParams.getSessionId()` prefers the session ID straight from the client.

If I understood correctly, for some reason session timeouts are not fully working somewhere under the hood.

Updating cookies on every action is more like handling the logic on our side, which might be fine but then it may be worth checking with Kimmo.

How does that sound?

gelbal: Agreed Pranas. I could only guess that Amplitude's reasoning is they want their SDK to be flexible when the client overwrites the `sessionId` -- so they don't apply the default timeout logic. Then perhaps there is a way to customize Amplitude behaviour. 

Can you please raise this point to Kimmo? 
It'd help to send the example sessions I linked here where we see how Amplitude doesn't apply timeout on the overwrites.

pranasziaukas: That's cool @gelbal, I'll message Kimmo to get a better understanding of the matter and let the analytics team know.

pranasziaukas:
@gelbal, @laziob neither Kimmo nor Amplitude support has replied to my inquiry about retaining the session timeout functionality after overriding the session ID.
I gave Kimmo another ping and hope to hear from someone soon.

gelbal:
Thanks for the update Pranas. Let's re-group about this tomorrow then. As we discussed, it's fine to pick a solution ourselves here and roll with it. So we can start collecting session data that would help us improve whatever solution we pick.

src/url-params.ts
laziob: @gelbal when you sent events to an old sessionId, did you tested using getSessionId() after timeout of a current session? or did you just set the session to be an old one from the USer Stream? Im wondering if the getSessionId() has a built in way to see if the session you are trying to retrieve timed out. This is because I saw that theres a method [here](https://amplitude.github.io/Amplitude-JavaScript/#amplitudeclientisnewsession) called isNewSession so it seems that it has some way to figure that out on its own.

gelbal: Good call @laziob.

`isNewSession` docs actually confused me. It says:
> Returns true if a new session was created during initialization, otherwise false.

We already checked the code and docs together on Wednesday and saw no `sessionId` parameter in the initialization config (like there is one for `deviceId`). Anyhow, looking at Amplitude JS source code, the `if` statement above this line explains when Amplitude sets `newSession` to true -- which is run during initialization indeed.
https://github.com/amplitude/Amplitude-JavaScript/blob/14fc693d56ff59a2a82b0eeadc4e7e6d3104026b/src/amplitude-client.js#L173

```js
if (!this._sessionId || !this._lastEventTime || now - this._lastEventTime > this.options.sessionTimeout)
```

So checking for `isNewSession` might be the solution indeed. Let's test it.

gelbal: To continue my train of thought: Actually, it'd not help in our case since there won't be `_lastEventTime` on the new component. 

We might as well apply the same logic ourselves on `analytics-client` -- which comes back to using cookie expiry date to track `_lastEventTime`.

Let's see how Amplitude support responds.

pranasziaukas:
@gelbal, as discussed, I don't think that storing session ID in a cookie provides any obvious benefit, so I'm removing it in a follow-up commit.

src/url-params.ts
```diff
@@ -16,11 +17,12 @@ const deviceIdSeparator = /\s*,\s*/;
 export class AnalyticsUrlParams {
 	private deviceIds: Set<string> = new Set();
 	private passedDeviceId: string | undefined;
+	private sessionId: number | null = null;
```
gelbal: As we did with `passedDeviceId`, shall we keep the types `number | undefined`? Any particular reason to assign it to `null`?

src/url-params.ts
```diff
@@ -121,6 +149,26 @@ export class AnalyticsUrlParams {
 		return `${URL_PARAM_DEVICE_ID}=${encodeURIComponent(ids.join(','))}`;
 	}
 
+	/**
+	 * @return part of the session ID query parameter string that can be appended to URLs
+	 */
+	getSessionIdQueryString(): string {
+		const id = this.getSessionId();
+		if (id === null) {
```
gelbal:
```suggestion
		if (id == null) {
```

So this returns true when checking `undefined == null`.
(in case we go ahead with removing the `null` from the type)

gelbal (commented):
Thanks for quickly taking action Pranas. This PR looks great. I added a minor comment on possibly changing the types.

src/url-params.ts
pranasziaukas: No reason :) Since the allowed type of `passedDeviceId` has changed to `undefined` we can do the same with `sessionId`.

gelbal (approved):
Great work Pranas. Let's ship it. 

As we've been discussing, it'd be great to closely monitor the behaviour of sessions linking in the wild when we update the components. We might need to add extra logic to have some control on when we set this.

PR #41, opened 2021-02-05
Changes on how device_id is treated on init

laziob:
With this changes the consumeUrlParams dont override the client
device_id with the passed one. Instead it returns the passed one
if exists and the clean queryParams string.
Then the amplitude Config is now allowed to recieve a device_id.
So then the client creation happens, if a device_id is passed on Config,
(which should be the one obtained from consumeUrlParams now),
it initiates the client with that device_id, and if not, it initiates
with no device_id, so it creates one.

Since this changes modifies some of the base logic of the
analytics-client, other components implementation should be
revisited to comply with these changes.

Change-type: Major

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,19 +11,42 @@ Currently the only final destination is Amplitude. We have plans to support more
 npm install --save analytics-client
 ```
 
-## Usage
 
-Tracking a page view. The following reports an event named `[EFP] Page View` with default properties.
+## Client Initialization
+The first thing to do is create and initialize an `analytics client`.
+
+Since most of balena products are maintained on different components/repos (ie Hub, Marketing Site, Cloud, etc) we need to make sure that
+when someone navigates between components we can keep identifying it as a same id. 
+
+So to achieve that we will need to:
 
 ```typescript
-import { createClient, createWebTracker } from 'analytics-client';
+import { AnalyticsUrlParams, createClient} from 'analytics-client';
 
+const urlParamsHandler = new AnalyticsUrlParams();
+
+// This will handle any passed device id (d_id) param on the URL if it exists, save it on a cookie and return a "clean" query string without the d_id param
+const newQuery = urlParamsHandler.consumeUrlParameters(window.location.search) ?? null;
+
+// Then if there was any passed device id we can obtain them like this.
+// passedDeviceId is either the first d_id param passed on the URL or null
+const passedDeviceId = urlParamsHandler.getPassedDeviceId()
+
+// Now that we checked for any passed device_id we can proceed with creating and initializing the client
 const client = createClient({
     endpoint: 'data.balena-cloud.com', // use 'data.balena-staging.com' for testing
     projectName: 'balena-project', // unique identifier that analytics-backend expects
     componentName: 'etcher-featured-project', // short unique name of component
     componentVersion: require('./package.json').version, // (optional) automated version reporting
-});
+    deviceId: passedDeviceId,
+})
+```
+
+## Usage
+Tracking a page view. The following reports an event named `[EFP] Page View` with default properties.
+
+```typescript
+import {createWebTracker } from 'analytics-client';
 
 createWebTracker(client, 'EFP') // 2nd parameter defines the prefix of event names
     .trackPageView();
--- a/src/client.ts
+++ b/src/client.ts
@@ -60,6 +60,9 @@ export interface Config {
 
 	/** Optional config for Amplitude client. */
 	amplitude?: Exclude<AmplitudeOverride, amplitude.Config>;
+
+	/** Optional device_id for Amplitude client. */
+	deviceId?: string;
 }
 
 interface AmplitudeOverride {
@@ -88,6 +91,10 @@ class DefaultClient implements Client {
 		amplConfig.includeReferrer = true;
 		amplConfig.includeUtm = true;
 		amplConfig.sameSiteCookie = 'Lax';
+
+		if (config.deviceId) {
+			amplConfig.deviceId = config.deviceId;
+		}
 
 		this.amplitudeInstance.init(config.projectName, undefined, amplConfig);
 		this.checkMixpanelUsage();
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -15,6 +15,7 @@ const deviceIdSeparator = /\s*,\s*/;
  */
 export class AnalyticsUrlParams {
 	private deviceIds: Set<string> = new Set();
+	private passedDeviceId: string | null;
 	private optOutRequsted: boolean = false;
 
 	constructor(private client?: Client) {
@@ -27,6 +28,8 @@ export class AnalyticsUrlParams {
 		currentDeviceId: string | null,
 	) {
 		const list = inputIdString ? inputIdString.split(deviceIdSeparator) : [];
+		this.passedDeviceId = list.length > 0 ? list[0] : null;
+
 		if (currentDeviceId) {
 			list.push(currentDeviceId);
 		}
@@ -96,6 +99,13 @@ export class AnalyticsUrlParams {
 		const res = new Set(this.deviceIds);
 		res.add(currentId);
 		return Array.from(res);
+	}
+
+	/**
+	 * @return an array of passed Device Ids via the URL param if there was any
+	 */
+	getPassedDeviceId() {
+		return this.passedDeviceId;
 	}
 
 	/**
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,7 @@
     "strictNullChecks": true,
     "moduleResolution": "node",
     "baseUrl": ".",
-    "keyofStringsOnly": true
+    "keyofStringsOnly": false
   },
   "include": [
     "index.ts",


src/url-params.ts
```diff
@@ -63,12 +63,12 @@ export class AnalyticsUrlParams {
...
-			if (this.client != null && newCurrentDeviceId != null) {
-				this.client.setDeviceId(newCurrentDeviceId);
-			}
+			//if (this.client != null && newCurrentDeviceId != null) {
+			//	this.client.setDeviceId(newCurrentDeviceId);
+			//}
 
 			params.delete(URL_PARAM_DEVICE_ID);
-			return params.toString();
+			return [passedDeviceId, params.toString()]; //Im passing `passedDeviceId` cause I dont know how to set the typescript return type in order to accept a list as returned by setDeviceIds()
```
pranasziaukas: > accept a list as returned by `setDeviceIds()`

`setDeviceIds` returns the first element of the array or `null`, its signature is
```typescript
private AnalyticsUrlParams.setDeviceIds(
    inputIdString: string | null | undefined,
    currentDeviceId: string | null): string | null
```
so an array/list cannot be obtained from it.

src/url-params.ts
```diff
@@ -49,7 +49,7 @@ export class AnalyticsUrlParams {
 	 * @param queryString URL query string to process
 	 * @return a modified query string that hides processed/consumed parameters if the query string should be modified
 	 */
-	consumeUrlParameters(queryString: string): string | null {
+	consumeUrlParameters(queryString: string): string[] | null {
```
pranasziaukas: You correctly labeled the commit as a major change because it breaks backwards compatibility.
Can we find a way around it?

src/client.ts
```diff
@@ -60,6 +60,9 @@ export interface Config {
 
 	/** Optional config for Amplitude client. */
 	amplitude?: Exclude<AmplitudeOverride, amplitude.Config>;
+
+	/** Optional device_id for Amplitude client. */
```
gelbal, line 64: Shall we make it possible to pass `userId` as well? 

It might help to initialize the analytics-client with a `userId` when the user is already logged in.

src/url-params.ts
gelbal: Indeed, I'd prefer not to introduce a breaking change if it's possible.

gelbal: Why do we need to pass a list here? 
To be clear, `consumeUrlParameters` uses the `setDeviceIds` function to parse and split string of `deviceId` values ([the relevant line](https://github.com/balena-io-modules/analytics-client/blob/3db19c67a74b61b7de33c6ddf2aa911d696a46c9/src/url-params.ts#L29))

src/url-params.ts
gelbal: `setDeviceIds` function stores the passed device identifiers under a cookie and we store all the values under the private property `deviceIds`. So why does this function need to return the parsed values back?

gelbal (commented):
Good idea to initialize the client with the passed `deviceId` already, Ezequiel. Then the changes in `url-params` are not clear to me. Why do we need these? 
I appreciate clarification.

src/url-params.ts
laziob: Right, my mistake. I thought setDeviceIds returned the whole array used to write the cookie values but indeed it returns only the first element of the array, which should be the passed value.

But the reason I changed it to return the parsed value was to then user it for Client init, without having to add the logic of reading the passed value from directly from the cookie later on.

src/url-params.ts
laziob: We dont, it stems from my confusion about setDeviceIds returning  the whole array instead of the first element of it.

src/client.ts
laziob: Sounds like a good idea!

src/url-params.ts
laziob: Right! Thanks Pranas, I missed that.

src/url-params.ts
laziob: Well the thing is that from what Ive understood of how its built, we could either leave it as it was (not returning the parsed passed device_id, just the query params) but removing the part that overrides the device_id on the client, which we now know its not the intended behaviour. But then for properly initializing the client we should add to it the logic of getting the passed device_id from the cookie that consumeUrlParameters wrote. At least thats another possible way than the one on this commit.

gelbal: @laziob I feel like I'm still missing the motivation behind updating `url-params` file. Let me paraphrase to see if I understand this. 

So we'd like to call `consumeUrlParameters` to parse the passed device identifier and use the first passed identifier as we initialize the client. Then we'd like to link all the device identifiers (passed and/or newly generated) -- which we retrieve from `allDeviceIds`.  

Do I get this right? What's the next problem that we need to solve here?

gelbal: Is the issue on the `setDeviceIds` call in the constructor of `AnalyticsUrlParams`?

laziob: @gelbal 
> So we'd like to call consumeUrlParameters to parse the passed device identifier and use the first passed identifier as we initialize the client.

Right, so far that is what I think should happen. Which in the end will result in less deviceIds per User since we wont be creating new ones on each init that later on get overriden by the passed one. 

> Then we'd like to link all the device identifiers (passed and/or newly generated) -- which we retrieve from allDeviceIds. 

I dont know exactly when this happens, but by looking at it, I dont think the behaviour of allDeviceIds and linkDevices is affected by this changes.

>Is the issue on the setDeviceIds call in the constructor of AnalyticsUrlParams?

Right, the issue per se in on [this lines](https://github.com/balena-io-modules/analytics-client/blob/ecc900e6699a854dead9a5483e87d00c03d11bfb/src/url-params.ts#L67) , which I commented. 

The other changes are just for returning the parsed passed device_id for using it on init, but that can be changed to any other method we find better or more convenient.

laziob: @gelbal @pranasziaukas I took a look again as to why I changed it to `string[]` and it was not because of expecting to recieve a list of device_ids from `setDeviceIds()` but because I want to return an array of strings that includes both the `passedDeviceId` and the `params.toString()` . The first one used for setting the client properly, the second one to build the clean URL with no `d_id` params.

Then again, I dont know if theres a better way on Typescript to handle this scenario

gelbal: I'd suggest using other functions to get these or implement new ones if necessary. So `AnalyticsUrlParams.allDeviceIds()` returns all the device identifiers as an array. The order is maintained, so the first identifier in the URL param is the first one in the array as well. The new generated identifiers are appended to the end of array.

gelbal:
```js
if (this.client != null && newCurrentDeviceId != null) {
	this.client.setDeviceId(newCurrentDeviceId);
}
```

Looking at the commented out code, I'd expect this to work as you explain Ezequiel: When we create `AnalyticsUrlParams` without a client set, this line shouldn't run. 

Shall we put aside the changes in `url-params` file and test your other changes? 
I feel like these are sufficient to solve the original issue here.

laziob: @gelbal if we put aside the changes to `url-params` it wont work, unless we do add on each component away to pass the passed `device_id`.

Right now the order in which the components are using the `analytics-client` is that they first create a client with `createClient()` which by defualt create a `device_id` and an `$identify` event with all the `init_x` properties, and then they run `AnlyticsUrlParams` functions such as `consumeUrlParams`. 

> When we create AnalyticsUrlParams without a client set, this line shouldn't run.

You are right that it wont run, but on the usual implementations like the one I described, theres always a client created. The first issue with the UI was that indeed we were not sending a `client` to the `AnalyticsUrlParams` call and thus we were not setting the proper `device_id` on the client. So following that flow, that line should not run only when no `newCurrentDeviceId` exists, but it should always expect a `client` because otherwise it breaks when there is in fact a `newCurrentDeviceId`.

So the order in which we avoid this kind of issue, is if we first check for a passed `device_id` with `consumeUrlParams`, we save that on cookies or anywhere that works, and only then, we go and create a client with either a passed `newCurrentDeviceId` or with none and we let the `createClient()` create a new `device_id`. 

Thats the reason I made the changes to the `url-params` file. To return both the `newCurrentDeviceId` needed for `createClient()` and the `params.toQueryString()` needed for the URL cleaning.

If I uncomment the commented code, we would still keep this logic of overriding the `device_id` of the client but it would actually never run because on the new needed order of things, `consumeUrlParams` should run before `createClient()` , at least for it to work as I think we intend to.

gelbal: @laziob, thanks for the clarifications. Indeed we need to make changes on the UI repo so the order of initialization is as you explain. Then I suggest doing such modifications on all clients before they start using the new `analytics-client` version.

My general point is: Let's build the ideal solution on `analytics-client` side and enforce the components to follow the expected initialization order. Editing the integration code in the components' repos is in our control. So we could modify the behaviour to follow our best practices. 

Let's assume that we are just rolling out the first version of `analytics-client` and there is no component using it yet, how would you implement this PR? How would you explain to integrate `analytics-client` to gracefully handle the cross domain navigation scenario?

To me it makes sense for a component to utilize `AnalyticsUrlParams` as first thing and parse any incoming parameter that might be required for the `analytics-client` initialization. 

If you agree that this solution would work, I suggest testing it together with the required changes on the UI side. Once we confirm that we solve the issues with the new solution, we would edit the component code to follow our new logic and roll out the new client together.

laziob: @gelbal Got it! 

Im positive that there might be a more elegant and optimal way to solve for the needed changes than the ones that I suggest here lol, but these changes follow that order of behaviour that I tried to illustrate on my previous comment. 

To go on par with this, I created [this other branch](https://github.com/balena-io/balena-ui/compare/eventLog_changes) on UI to see what it would look like to use this version of `analytics-client` . I think it would help if you could give it a look to see if there is something I missed that could break the implementation. But basically I did exactly that: Change the order in which we run the different functions to comply with these changes.

But either way I will keep on working on this branch to see if  I can come up with a more readable, user friendly way to implement this new logic

gelbal: Cool. Please create a UI PR with that branch Ezequiel. So we could all start commenting on the code and work together towards a solution. I could also start testing the proposed solution in other components like Hub and the marketing website.

I find the current solution pretty nice actually. I will think further if we could come up with a more elegant solution.

pranasziaukas: I wrote this while reviewing another PR but it's just as relevant here:

Maybe deconstructing tuples can help, like
```typescript
const [deviceId, newQuery] = consumeUrlParameters(window.location.search);
```
as described in https://www.carlrippon.com/strongly-typed-destructuring-and-rest-parameters/

Or just use two separate method calls :)

laziob: @gelbal @pranasziaukas  I was looking into change how the UI init works by using `AnalyticsUrlParams.allDeviceIds()` so as to keep `consumeUrlParams()` return value as before. But I stumble with the following issue:

Many of the methods like `allDeviceIds()` require that the `client` has already been passed to `AnalyticsUrlParams`. But since we are trying to init the `client`, we cant use this methods since there is no `client` to pass to them yet.

So basically my changes to `consumeUrlParams()` are not very good since Im just hardcoding it to bypass the need of a `client` but all the other methods on the class needs it.

What do you propose to do here?

src/client.ts
laziob: Actually we cant, `user_id` is not a passable parameter to the `.init` method. The passable parameters are listed [here](https://amplitude.github.io/Amplitude-JavaScript/Options/#options)

src/url-params.ts
gelbal: @laziob I don't think we have an issue here. Here is the `client` reference in [`allDeviceIds()`](https://github.com/balena-io-modules/analytics-client/blob/3db19c67a74b61b7de33c6ddf2aa911d696a46c9/src/url-params.ts#L91):
```js
const currentId = this.client != null ? this.client.deviceId() : null;
```

So it should be fine if there is no `client` set. As far as I see `AnalyticsUrlParams` is designed to work without a `client` set. If we find a breaking case, let's update it for sure.

I suggest addressing the feedback on the UI PR and update the code not to depend on `consumeUrlParams`'s output to get `deviceId`. Instead `allDeviceIds()[0]` should return either the first `d_id` in URL params or null. 

Let's test these together with the changes here (putting aside the changes in `url-params`). I believe initializing the `client` with `deviceId` is a sufficient solution. Then once we have `client` initialized, we'll set it in `AnalyticsUrlParams`. I will test this solution on another component.

laziob: @gelbal you were right, I made some changes on the UI repo to handle the `passedDeviceId` with `allDeviceIds()[0]` and it seems to work, both when a device id is passed and when its not. I already made the changes on a new commit on the UI PR. Please let me know what you think and if its okey I can revert the changes to `url-params-handler` here on a new commit.

gelbal: Nice. Looks great Ezequiel. Yes, please revert the changes in `url-params` here so we could ship this PR as well. Consecutively, we are not breaking backwards compatibility so we could roll the new version out at our own pace.

pranasziaukas: Just to recap and make sure we're all on the same page, let's:
1. Initialize `AnalyticsUrlParams` without a client.
1. Use `AnalyticsUrlParams.consumeUrlParams` as needed.
1. Use `AnalyticsUrlParams.allDeviceIds` to get the relevant `deviceId`.
1. Initialize the `Client` with the `deviceId`.
1. Inject the client using `AnalyticsUrlParams.setClient`.

Voila!

laziob: @pranasziaukas  yes I think thats about right, although I havent used de `AnalyticsUrlParams.setClient` on the UI changes.  I think it should run as `urlParamsHandler.setClient(analyticsClient)` right above [this line](https://github.com/balena-io/balena-ui/blob/9804d2216e7af2ebb4c4acf7abc4255f5a7e3efd/src/services/eventLog.ts#L36) but then in order to use it later on, the `initAnalyticsClient()` should return the `urlParamsHandler` instead of the `analyticsClient` or both?

laziob: > please revert the changes in url-params here so we could ship this PR as well

@gelbal done!

src/client.ts
gelbal: Right. Seems like we need to use `identify` for user identification. Thank for checking!

src/url-params.ts
gelbal: I like that summary @pranasziaukas noted here. @laziob do you mind updating the docs (Readme) to reflect this? 

So we explain the current practice outside of this PR as well.

gelbal (dismissed):
Let's ship it! Thanks for coming up with the idea and testing this Ezequiel. Good stuff.

src/url-params.ts
laziob: Sure, I'll update it asap.

gelbal:
Thanks for the documentation edit Ezequiel. This is good to ship in my opinion. Please remove the draft tag and re-ask for review or comment `@balena-ci I self-certify!` in the repo to merge this by yourself already.

As I asked in the linked UI PR, feel free to address such a change if you also agree that it helps:
Shall we create a convenience method in `url-params` for this call?
```js
const passedDeviceId = urlParamsHandler.allDeviceIds()[0];
```

Something like `urlParamsHandler.getInitialDeviceId()` might convey the logic here better.

laziob:
@gelbal 

> Shall we create a convenience method in `url-params` for this call?
> 
> ```js
> const passedDeviceId = urlParamsHandler.allDeviceIds()[0];
> ```
> 
> Something like `urlParamsHandler.getInitialDeviceId()` might convey the logic here better.

mmm I found "initial" to be a somewhat confusing term. We are referring to the first element of the array of `devices ids` but I dont know if that is exactly a "inital" `device_id`. Also I cant think of any other use case where someone would need to get that first element other the one we are using, at least not yet. 

I do think that writing a convenience method for getting the `passedDeviceId` could be more clear for the End User though. 

Something like: 
`urlParamsHandler.getPassedDeviceId()` or `urlParamsHandler.getCrossTrackingDeviceId()`

What do you think?

gelbal:
`urlParamsHandler.getPassedDeviceId()`

I like this @laziob :+1:

pranasziaukas:
@laziob I think at least one of the CI problems has to do with linting
```
Error: File src/client.ts hasn't been formatted with prettier
Error: File src/url-params.ts hasn't been formatted with prettier
```
You could try running `npm run prettify` to get the styling in place (although I thought the linter runs automatically on commit courtesy of [husky](https://www.npmjs.com/package/husky)).

laziob (commented):
@gelbal @pranasziaukas  here are the changes to get passed device ids. What I did was to use some of the work that `setDeviceIds()` does to get only the passed device ids before they are joined with any existing device ids and store them on a separate const at the class level. 

Then the `getPassedDeviceIds()` just checks for this class const and if it has some non null value it returns undefined susch as allDeviceIds() works.

please let me know if you are ok with this. Thanks!

src/url-params.ts
```diff
@@ -27,6 +28,8 @@ export class AnalyticsUrlParams {
 		currentDeviceId: string | null,
 	) {
 		const list = inputIdString ? inputIdString.split(deviceIdSeparator) : [];
+		this.passedDeviceIds = new Set(Array.from(list));
```
pranasziaukas: `list` is already an array (as per 1 line above) so not sure if the extra conversion is of any need?

README.md
```diff
@@ -11,19 +11,42 @@ Currently the only final destination is Amplitude. We have plans to support more
...
+import { AnalyticsUrlParams, createClient} from 'analytics-client';
+
+const urlParamsHandler = new AnalyticsUrlParams();
+
+// This will handle any passed device id (d_id) param on the URL if it exists, save it on a cookie and return a "clean" query string without the d_id param
+const newQuery = urlParamsHandler.consumeUrlParameters(window.location.search) ?? null;
+
+// Then if there was any passed device id we can obtain them like this.
+// passedDeviceId is either the first d_id param passed on the URL or null, so getPassedDeviceIds()[0] works even when no d_id is sent on the URL.
+const passedDeviceId = urlParamsHandler.getPassedDeviceIds()[0]
```
pranasziaukas: Do we have any suspected usages of the multitude of passed device IDs?
At this point I expected a `getPassedDeviceId` (singular) method to handle only one ID as discussed in previous comments :)

src/url-params.ts
```diff
@@ -98,6 +101,15 @@ export class AnalyticsUrlParams {
 		return Array.from(res);
 	}
 
+	/**
+	 * @return an array of passed Device Ids via the URL param if there was any
+	 */
+	getPassedDeviceIds() {
+		const passedIds = this.passedDeviceIds;
+		const res = new Set(passedIds);
+		return Array.from(res);
```
pranasziaukas: Maybe it would be better to run the logic once in `setDeviceIds` instead of on every retrieval?
Furthermore, I'm confused about the repeated casting between `Set` and `Array` types - is there any meaning to it?

pranasziaukas (commented):
Thanks Ezequiel. I believe we're moving in the right direction but there're still some things I'd like to clarify before deploying.

README.md
gelbal: For sure it could happen. A user might come to Hub from various sources, marketing website, EFP, blog. Imagine someone using Etcher visiting the Hub for the first time via EFP and ending the session without further navigation or signup event. Then the same user later visits our blog from other means than EFP and navigates to Hub from the blog. On the Hub we will end up with multiple device identifiers to pass in the URL when the user navigates to the dashboard.

gelbal: On a related note, it'd be great to edit the Readme and this specific comment as we implement the new convenience method.

src/url-params.ts
gelbal: Agreed with Pranas. On a different note, I suggest refactoring this function to return a single device ID, instead of an array. 

We already have `allDeviceIds()` that returns all the device identifiers. Is there a need to return specifically the passed device identifiers?

README.md
pranasziaukas: > On the Hub we will end up with multiple device identifiers to pass in the URL when the user navigates to the dashboard.

I'd use `allDeviceIds` method for this instead of `getPassedDeviceIds`, but looking at the comment below I think you mean the same thing :)

src/url-params.ts
pranasziaukas: > We already have `allDeviceIds()` that returns all the device identifiers. Is there a need to return specifically the passed device identifiers?

This was exactly my confusion in the comment above.

README.md
gelbal: Oh now I get your first comment. Yes we are on the same page : )

laziob: Right, there was no use case I could think about for all the passed ids, but I proceeded this way to differentiate between Ids that are passed from components from those that could be generated by the component in case it was needed afterwards for some reason.

thats why the method gets all the passed device ids and not only one. If we decide to build a function to get just one passed id, should it always be the first one from the array? Or could it be some other one for some reason? Since I wasnt sure, I though this function at least gives us the chance to decide afterwards.

But let me know and I could change it or create a new one for just the "last device id passed" (first one on the array)

src/url-params.ts
laziob: > We already have allDeviceIds() that returns all the device identifiers. Is there a need to return specifically the passed device identifiers?

My reasoning was that for some visits,  the `allDeviceIds()` function gets the total of devices already stored in the client plus those passed via the URL, but there is no way of distinguishing between those once they are all in the same array. So this way, in case there is a need later on, we can differentiate between those already stored by the client and those that were obtain by the d_id param.

gelbal: No need to support an unknown use case now @laziob. We could add that when the need arises. Besides, we are now designing the device linking logic to always prefer the 1st passed device ID over everything else. So it's perfectly fine to offer an opinionated method here (which always returns the 1st passed value).

laziob: @gelbal @pranasziaukas 

I understand. I just commited the new changes. Please let me know how it looks now.

README.md
```diff
@@ -11,19 +11,42 @@ Currently the only final destination is Amplitude. We have plans to support more
...
 
+// Now that we checked for any passed device_id we can proceed with creating and initializing the client
 const client = createClient({
     endpoint: 'data.balena-cloud.com', // use 'data.balena-staging.com' for testing
     projectName: 'balena-project', // unique identifier that analytics-backend expects
     componentName: 'etcher-featured-project', // short unique name of component
     componentVersion: require('./package.json').version, // (optional) automated version reporting
-});
+	deviceId: passedDeviceId,
+	})
```
pranasziaukas: These two lines are misaligned with the rest (tab instead of spaces?)

src/url-params.ts
```diff
@@ -98,6 +101,15 @@ export class AnalyticsUrlParams {
 		return Array.from(res);
 	}
 
+	/**
+	 * @return an array of passed Device Ids via the URL param if there was any
+	 */
+	getPassedDeviceIds() {
```
pranasziaukas: I believe this is now `getPassedDeviceId` (singular)

README.md
laziob: Mm thats odd, I dont see it like that on local. Maybe something went off with prettify? Let me check

src/url-params.ts
laziob: Yes you are right, let me fix it asap

README.md
pranasziaukas: I don't think prettify covers markdown files, it's up to the editor

src/url-params.ts
```diff
@@ -27,6 +28,8 @@ export class AnalyticsUrlParams {
 		currentDeviceId: string | null,
 	) {
 		const list = inputIdString ? inputIdString.split(deviceIdSeparator) : [];
+		this.passedDeviceIds = list;
```
gelbal: Better to do a deep copy here Ezequiel. The lines just below might modify `list` and thus `passedDeviceIds` by pushing `currentDeviceId`.

gelbal:
@laziob before we merge this, can you please squash your commits?

src/url-params.ts
pranasziaukas: Maybe we want to just take the first element `list.length > 0 ? list[0] : null` using the same reasoning of no need to support unknown use cases?

laziob: You mean something like: 

`const list = inputIdString ? inputIdString.split(deviceIdSeparator) : [];`
`this.passedDeviceIds = inputIdString ? inputIdString.split(deviceIdSeparator) : []; `

?

README.md
```diff
@@ -11,19 +11,42 @@ Currently the only final destination is Amplitude. We have plans to support more
...
 
+// Now that we checked for any passed device_id we can proceed with creating and initializing the client
 const client = createClient({
     endpoint: 'data.balena-cloud.com', // use 'data.balena-staging.com' for testing
     projectName: 'balena-project', // unique identifier that analytics-backend expects
     componentName: 'etcher-featured-project', // short unique name of component
     componentVersion: require('./package.json').version, // (optional) automated version reporting
-});
+    deviceId: passedDeviceId,
+	})
```
pranasziaukas: This line is still misaligned, sorry Ezequiel :)

src/url-params.ts
laziob: @pranas sorry just saw your comment that was made at the same time. I see what you mean, let me fix that.

README.md
laziob: :( I dont know why though. I'll keep looking

README.md
```diff
@@ -11,19 +11,42 @@ Currently the only final destination is Amplitude. We have plans to support more
...
 
+// Now that we checked for any passed device_id we can proceed with creating and initializing the client
 const client = createClient({
     endpoint: 'data.balena-cloud.com', // use 'data.balena-staging.com' for testing
     projectName: 'balena-project', // unique identifier that analytics-backend expects
     componentName: 'etcher-featured-project', // short unique name of component
     componentVersion: require('./package.json').version, // (optional) automated version reporting
-});
+    deviceId: passedDeviceId,
+    })
```
pranasziaukas:
```suggestion
})
```
I should've made a change suggestion right away - sorry about that

laziob:
@gelbal  I tried to squash the commits but then it asked me for a merge. Please let me know if the squash worked as intended or if I should somehow squash it again. Thanks!

Also @pranasziaukas I think the ReadMe now its properly indented.

src/url-params.ts
```diff
@@ -15,6 +15,7 @@ const deviceIdSeparator = /\s*,\s*/;
  */
 export class AnalyticsUrlParams {
 	private deviceIds: Set<string> = new Set();
+	private passedDeviceId: string | null = '';
```
gelbal: Can you please not set this to empty string by default Ezequiel?
```suggestion
	private passedDeviceId: string | null;
```

Just to prevent any potential issues down the line.

gelbal:
@laziob it doesn't seem like squash worked. I see 12 commits in the PR now. Let me know if you'd like to do it together.

laziob:
@gelbal  le me try one more time and if I cant make it I'll let you know so we can do it together

laziob:
@gelbal I somehow got it to 3 commits haha. Still coudnt squashed them all to 1 but let me know if this works

gelbal:
@laziob we are almost there :smile: Can you please squash all into 1 commit? 
These 3 commits are undoing each others edits and there is no need to write these back-and-forth changes to our git history.

How do you use git? 
I use the commandline and I'd do such squash with the command `git rebase -i HEAD~3` and interactively edit the commit messages to merge everything.

gelbal:
On a related note. If the `git push` is the problem, use force :laughing: 

I use this relatively safer force push: `git push --force-with-lease`

laziob:
@gelbal Now I think I done it properly!

gelbal (approved):
Sweet. Thanks for pushing this through Ezequiel. Good stuff. Let's deploy this everywhere!

laziob:
@gelbal I will now update the UI PR to be in line with this changes. Where can I get the new version for this so as to also update it on that repo?

pranasziaukas:
Uhm, we haven't changed the `Change-type: Major` so the versionbot released it as v1.0.0, which is misleading. Shall we rollback and resubmit as a minor change?

gelbal:
Good point Pranas. Then in a sense this is a major update as we significantly refactored how we initialize the Amplitude client in almost all of the components utilizing `analytics-client` now. So I'm fine with keeping it 1.0.0.

laziob:
But isnt it a Major change since it implies to change how the other
components use the analytics-client too? Where is the line of what is
considered a Major and Minor change?

gelbal:
> Where can I get the new version for this so as to also update it on that repo?

@laziob I don't how long the lag would be. I hope by the time you update the UI PR, you'll be able to retrieve the new version. So please update `package.json` entry for `analytics-client` to point to 1.0.0 or higher. Let's see if npm would be able to retrieve it fine already.

pranasziaukas:
Not urgent, but I'd suggest adding at least some unit tests alongside new methods in the future.

src/url-params.ts
```diff
@@ -98,6 +101,13 @@ export class AnalyticsUrlParams {
 		return Array.from(res);
 	}
 
+	/**
+	 * @return an array of passed Device Ids via the URL param if there was any
```
pranasziaukas, line 105: The description is from `getPassedDeviceIds` and no longer accurate

PR #42, opened 2021-02-16
Add trackNavigationClick

gelbal:
Change-type: minor
Signed-off-by: Fırat Gelbal firat@balena.io

### DIFF
--- a/src/web.ts
+++ b/src/web.ts
@@ -5,6 +5,7 @@ import { Client, Properties } from './client';
  */
 export interface WebTracker {
 	track(name: string, props?: Properties): void;
+	trackNavigationClick(href: string, props?: Properties): void;
 	trackPageView(name?: string): void;
 }
 
@@ -12,6 +13,15 @@ export function createWebTracker(client: Client, prefix?: string): WebTracker {
 	return {
 		track(name: string, props?: Properties) {
 			client.track(`${prefix ? `[${prefix}] ` : ''}${name}`, props);
+		},
+
+		trackNavigationClick(href: string, props?: Properties) {
+			this.track('Navigation Click', {
+				...props,
+				href,
+				current_url: window.location.href,
+				current_url_path: window.location.pathname,
+			});
 		},
 
 		trackPageView(name?: string) {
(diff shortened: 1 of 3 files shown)


src/web.ts
```diff
@@ -14,6 +15,14 @@ export function createWebTracker(client: Client, prefix?: string): WebTracker {
 			client.track(`${prefix ? `[${prefix}] ` : ''}${name}`, props);
 		},
 
+		trackNavigationClick(props?: Properties) {
+			this.track('Navigation Click', {
+				...props,
+				current_url: window.location.href,
+				current_url_path: window.location.pathname,
```
laziob, line 23: @gelbal  If its a Navigation Click, shouldnt we add also the "a href" value too just in case? ie: The [Hub] Project Navigation click already has it, and it could be useful to ask for it on all Navigation Clicks I believe.

gelbal: You mean the `href` value, right? Yeah, I wanted to capture it automatically but I don't know if there is a clean & straightforward way to capture it on `analytics-client` (like we capture the `window.location` values). Is there?

gelbal: Thinking it further now, if we are not able to capture `href` automatically here, it'd be better to require the `href` property in this function. So the component maintainers know that they have to pass the destination of the navigation.

laziob: Lol I just edited my comment because since I wrote the href tag explicitly it got itnerpreted as link haha. Yes, I mean the href of any button or "<a>" element. 

And I cant think of a way to get that since its to tied up to the front, but 
> t'd be better to require the href property in this function

this sounds perfect for this case.

gelbal: Cool. I made that change now to require `href` parameter.

pranasziaukas: Looks like a nice quality of life improvement. Would we like to mention it in the README as well?

gelbal: Sure, why not. I edited Readme. Please take another look Pranas.

pranasziaukas (approved):
LGTM - concise changes with accompanying tests and updated documentation.

pranasziaukas:
As I see, this PR is also suffering from the `"keyofStringsOnly": false` issue :)

PR #46, opened 2021-02-24
Fix `optOutRequested` typo

pranasziaukas:
Change-type: patch
Signed-off-by: Pranas Ziaukas <pranas@balena.io>


pranasziaukas:
@balena-ci I self-certify!

PR #47, opened 2021-02-25
Changed getQueryString to validate domains

laziob:
Since we only want to append device id and session id when needed,
that is, when navigating cross domain or cross component, now the
getQueryString function checks if there is a passed destination
hostname and if so it compares its main domain to the main domain of
the actual URL the user is on. When a difference between the two exists,
or there was no destination hostname passed it returns both
d_id and s_id params. Otherwise if these 2 are the same, it returns
an empty string.

Change-type: Minor

### DIFF
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -163,10 +163,37 @@ export class AnalyticsUrlParams {
 	/**
 	 * @return full query parameter string that can be appended to URLs
 	 */
-	getQueryString(): string {
-		return [this.getDeviceIdsQueryString(), this.getSessionIdQueryString()]
-			.filter(x => x)
-			.join('&');
+	getQueryString(destinationUrl?: URL, currentUrl?: URL): string {
+		// this regex is based on the assumption that we wont be using TLDs longer than 3 characters. If we do, it will break
+		// the logic and take that longer TLD as the main domain, for example hub.balena.edge.io -> edge.io
+		const regex = /([a-zA-Z0-9-]+)(\.[a-zA-Z]{2,3})?(\.[a-zA-Z]+$)/g;
+
+		let actualDomainMatch;
+		if (currentUrl) {
+			actualDomainMatch = currentUrl.hostname.match(regex);
+		} else if (typeof window !== 'undefined') {
+			actualDomainMatch = window.location.hostname.match(regex);
+		} else {
+			actualDomainMatch = null;
+		}
+
+		const destinationDomainMatch = destinationUrl
+			? destinationUrl.hostname.match(regex)
+			: null;
+
+		const actualDomain = actualDomainMatch
+			? actualDomainMatch.toString()
+			: null;
+		const destinationDomain = destinationDomainMatch
+			? destinationDomainMatch.toString()
+			: null;
+
+		if (!destinationDomain || actualDomain !== destinationDomain) {
+			return [this.getDeviceIdsQueryString(), this.getSessionIdQueryString()]
+				.filter(x => x)
+				.join('&');
+		}
+		return '';
 	}
 
 	/**
--- a/test/url-params.test.ts
+++ b/test/url-params.test.ts
@@ -102,6 +102,43 @@ test('not changing current URL', () => {
 	expect(newQuery).toBeNull();
 });
 
+test('parsing and matching destination and actual URL to regex', () => {
+	const urlParams = new AnalyticsUrlParams();
+	expect(urlParams.getQueryString()).toBe('');
+
+	urlParams.consumeUrlParameters('s_id=123&other=value');
+	expect(urlParams.getQueryString()).toBe('s_id=123');
+
+	// Case when not passing any destination or actual URL
+	urlParams.consumeUrlParameters('d_id=d1&other=value');
+	expect(urlParams.getQueryString()).toBe('d_id=d1&s_id=123');
+
+	// Case when passing matching destination and actual URL while a d_id and s_id exist.
+	expect(
+		urlParams.getQueryString(
+			new URL('https://test.domain.io'),
+			new URL('https://domain.io'),
+		),
+	).toBe('');
+
+	// Case when passing none matching destination and actual URL while a d_id and s_id exist.
+	expect(
+		urlParams.getQueryString(
+			new URL('https://test.domain.io'),
+			new URL('https://otherdomain.com'),
+		),
+	).toBe('d_id=d1&s_id=123');
+
+	// Case when passing destination URL that violates the TLD regex assumption. Here it will result in matching URL "edge.io", when in fact
+	// It should be treated as different URLs and return the d_id and s_id params
+	expect(
+		urlParams.getQueryString(
+			new URL('https://test.domain.edge.io'),
+			new URL('https://domain2.edge.io'),
+		),
+	).toBe('');
+});
+
 interface AnalyticsMock {
 	setDeviceIdParams: string | null;
 	deviceIdRetrieved: boolean;


src/url-params.ts
```diff
@@ -163,10 +163,20 @@ export class AnalyticsUrlParams {
 	/**
 	 * @return full query parameter string that can be appended to URLs
 	 */
-	getQueryString(): string {
-		return [this.getDeviceIdsQueryString(), this.getSessionIdQueryString()]
-			.filter(x => x)
-			.join('&');
+	getQueryString(destinationHostname?: string): string {
```
pranasziaukas: Just thinking about the use case, what do we expect from the UI code?
If we're preparing to navigate somewhere maybe the full URL is already at hand?

What I want to say, it would be good to not require any additional (verbose) operations or the UI side, trying to figure out if extracting the hostname would count as one.

pranasziaukas (commented):
Left a small point for discussion.

Other than that, I think at least some unit test coverage would be very nice.

src/url-params.ts
gelbal: Agreed with Pranas. It's simpler to expect the full destination URL here instead of hostname. I'd suggest to make this function work with any valid URL string.

laziob: Right, I wrote it like this because Im not familiar to how each component handles this function. Hub for example was transforming a url string to a `URL` type for the base url, so If I wanted to use that, I would have to switch the type of the argument to URL. Then maybe marketing site uses just `window.location.href` and then It would need to transform it to a URL type to pass it and so. So I though that to make it explicit for the component to pass just the hostname would be best but Im not sure.

src/url-params.ts
```diff
@@ -163,10 +163,20 @@ export class AnalyticsUrlParams {
...
+		const regex = /([a-zA-Z0-9-]+)(\.[a-zA-Z]{2,5})?(\.[a-zA-Z]+$)/g;
```
gelbal: Good stuff with this regex here. It works with all cases I could imagine on balena websites. Then it returns `test.test.io` as is.

```js
(new URL('https://test.test.io')).hostname.match(regex) -> test.test.io
```

Can you please expand the reasoning behind the numbers in `{2,5}`? 

Setting it to `{2,3}` makes my test case return `test.io` and doesn't break other tests. That's why I'd like to understand this 5 number.

src/url-params.ts
pranasziaukas: Yeah, I can see how it depends on our component implementations.

However, even in different use cases, I'd expect these things:
- If URLs are at hand then maybe they can be casted easily, i.e. `String(window.location.href)` and so
- Usage to converge over time, i.e. components to start using similar means to navigate

gelbal: Since this is a new method and we'd like to roll it out ourselves, it's fine for us to dictate the format and type of the parameter here. We will update the relevant places on components to follow the practice we set. Then I don't have a strong opinion on expecting a URL object or URL string here. Since this is a Typescript project, maybe it's better to strictly use URL type.

https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.url.html

src/url-params.ts
```diff
@@ -163,10 +163,20 @@ export class AnalyticsUrlParams {
...
+		const actualDomain = window.location.hostname.match(regex);
```
gelbal: Shall we support `currentUrl` as parameter as well? Is it safe to assume that `UrlParams` class always has access to `window`?

I feel like there could be a case that a component wants to ask for URL params for other places beyond `window.location`.

src/url-params.ts
```diff
@@ -163,10 +163,20 @@ export class AnalyticsUrlParams {
...
+		const destinationDomain = destinationHostname
+			? destinationHostname.match(regex)
+			: undefined;
+
+		if (actualDomain !== destinationDomain) {
```
gelbal: `String.match(regex)` returns Array or null. So we need to update the comparison here.

src/url-params.ts
```diff
@@ -163,10 +163,20 @@ export class AnalyticsUrlParams {
...
+			return [this.getDeviceIdsQueryString(), this.getSessionIdQueryString()]
+				.filter(x => x)
+				.join('&');
+		} else {
+			return '';
+		}
```
gelbal:
```suggestion
		return '';
```

I believe this is the standard we've been following for default returns (instead of explicit `else`).

gelbal (commented):
Good start Ezequiel. I added couple of comments and questions. 

I agree with Pranas to have unit test coverage here. Additionally, it'd be great to update the example on Readme and/or add one more example.

Finally, shall we wrap this logic in a new function to offer it under `getDeviceIdsQueryString` and `getSessionIdQueryString` individually? Do we expect the components to only pass `deviceId` and not `sessionId` in the URL ever? (or vice versa)

src/url-params.ts
laziob: @gelbal @pranasziaukas  just to be clear, what I meant is that, at least on Hub for example, they dont hande the URL as a string, but they handle it as an URL object. So I have to use any of the string attributes of it, be it href, hostname, or whatever. While on Docs for example [on this line](https://github.com/balena-io/docs/blob/3183fdc0f98ea21487b86635dc0f693183461fa2/templates/_track.html#L38) they handle everything on the return explicitly for links that send to dashboard only, there is no logic for all external links, so the maintainer would have to either explicitly set an argument, either the href, hostname, whatever, or send the URL object, but not both.

laziob: So if we built getQueryStrings to expect URL objects, then each maintainer would have to make the change to take the href and build the URL object. If we leave it as to expect a string, it would need to explicitly send href, hostname or whatever. Or I could write the function to expect both either a URL object or a string and proceed with the respective logic whatever the case and then let each maintainer to choose to sue whatever suits him best. Should we aiming for the last scenario?

laziob: Now its prepared to recieve both destinationURL and currentURL, both as URL objects and parse them for the regex.

src/url-params.ts
laziob: I was using 5 because it was to handle those unusual TDL like ".forum", ".audio" etc. But I didnt noticed that it could missinterpret short domain names like "test". I switched to 3 now assuming we wont be using weird long TDLs for the moment.  Because otherwise, things like "hub.balena.edge.io" where hub is a subdomain, balena a domain, and .edge.io TLD , it matches only "edge.io"

src/url-params.ts
laziob: Just in case this happens I added the currentURL parameter to the function and the logic to parse it in case there is one, otherwise it tries to use window.location.hostname

src/url-params.ts
laziob: Yeap, Im using toString() now with some extra validation to avoid using it on a null entity.

src/url-params.ts
laziob: Fixed!

src/url-params.ts
```diff
@@ -111,6 +111,15 @@ export class AnalyticsUrlParams {
 		this.client = client;
 	}
 
+	/**
+	 *
+	 * @param client
+	 * sets the class client value to whatever client is passed as the function argument
+	 */
+	declareClient(client: Client) {
```
pranasziaukas: How is this related to `getQueryString`?

I get it is related to the initialization of the tracking client but might've been better off as a separate PR. Just nitpicking.

src/url-params.ts
```diff
@@ -163,10 +172,28 @@ export class AnalyticsUrlParams {
...
 	 */
-	getQueryString(): string {
-		return [this.getDeviceIdsQueryString(), this.getSessionIdQueryString()]
-			.filter(x => x)
-			.join('&');
+	getQueryString(destinationUrl?: URL, currentUrl?: URL): string {
+		const regex = /([a-zA-Z0-9-]+)(\.[a-zA-Z]{2,3})?(\.[a-zA-Z]+$)/g;
+		const actualDomainMatch = currentUrl
+			? currentUrl.hostname.match(regex)
+			: window.location.hostname.match(regex);
```
pranasziaukas: What happens if the `window` object is not accessible?
Is it going to be `undefined`, throw an exception or do something else? Just trying to cover the logical cases here.

src/url-params.ts
gelbal: I don't think it's nitpicking. I'd also suggest to split such work to a separate PR. We'd need to test the impact of using such a new method a bit further and it could use some more discussion. There is probably no need to hold the original changes here waiting for such further discussion and tests around `declareClient`.

src/url-params.ts
gelbal: Got it. 

> assuming we wont be using weird long TDLs

That's fair. Can you please comment this assumption in code Ezequiel?

Additionally, it'd be great to have unit tests here that shows the expected values for such transformation. Perhaps one of the tests could be `hub.balena.edge.io -> edge.io` to illustrate your assumption and the expected behavior.

gelbal:
Hey @iamsolankiamit here is the PR for the changes around URL parameter generation we've been discussing on FD. Your review is appreciated.

src/url-params.ts
laziob: My bad, I had added that to be able to test the session_id being passed but was indeed intended for a separate PR as discussed on FD, I'll remove it on the next commit

src/url-params.ts
laziob: I added the assumption on both the code and on the test to explicitly state what should happen and why

src/url-params.ts
laziob: @pranasziaukas now I wrote it differently for it to check also if window object exists

gelbal:
Good stuff with the tests @laziob. It'd be neat to mention why all the consecutive tests in the same run returns the previously set `d_id` and `s_id` values. Furthermore, it might help to split some of the tests so they don't all share the same context.

gelbal:
One more discussion point before we merge this: Shall we maintain a list of balena domains as a configuration option? 
I'd love to hear your thoughts @laziob @pranasziaukas @sradevski @iamsolankiamit. 

The main advantage of knowing the domains to be tracked is we could configure `analytics-client` not to return query parameters on external links. In the current implementation, the component maintainer would get query params for links like `twitter.com`, `facebook.com`. Then the tradeoff is the maintenance of such list. Embedding it on `analytics-client` seems like a no go from product builder perspective. Maybe it could be part of that secret JSON mapping that we use to retrieve Amplitude tokens. Still, it's a burden to maintain.
Alternatively, we could ask the component maintainers to only make URL Params call on internal links and not on external links. Else, `analytics-client` will append `d_id` and `s_id` params on twitter / facebook links.

laziob:
@gelbal I used the same run because that run covered , at a general level, all posible scenarios, which are either no param passed, only one param passed, or both params passed. And at each instance it has its expected value. Then other possible changes are differences in values, but those functions dont make any checks on what the value should be, so I didnt see the case on tsting with values different than the '123' used there. 

If there is some other context that you have in mind please let me know so I can build the proper test for it.

Regarding the domains, Im with you that keeping some kind of registry might be a burden further along the road. At the moment, for example the Hub does exactly what you mentioned. It has different classes I think and depnding if the link is an internal or external one the apply different onClick functions. While Blog I think it was, handles the onClick directly on the button with the specific needed URL. 

I really cant say whats the best practice here, so looking forward to hear the others opinions!

pranasziaukas:
> Shall we maintain a list of balena domains as a configuration option?

Personally, I'm not a huge fan of maintaining such lists/mappings because it is destined to fail eventually.

> Alternatively, we could ask the component maintainers to only make URL Params call on internal links and not on external links.

I think this makes sense. Right now, the only real reason to use `getQueryString` (that I'm aware of) is to generate context to pass to further destinations - at that point, the conscious decision to pass the parameters is already made, and the decisionmaking might as well include the internal-external domains.

In other words, we should not blindly concatenate `destinationUrl + urlParams.getQueryString(destinationUrl)` anyway because that can easily result in unexpected side effects. Or should we? :)

src/url-params.ts
```diff
@@ -163,10 +163,37 @@ export class AnalyticsUrlParams {
...
-			.join('&');
+	getQueryString(destinationUrl?: URL, currentUrl?: URL): string {
+		// this regex is based on the assumption that we wont be using TLDs longer than 3 characters. If we do, it will break
+		// the logic and take that longer TLD as the main domain, for example hub.balena.edge.io -> edge.io
+		const regex = /([a-zA-Z0-9-]+)(\.[a-zA-Z]{2,3})?(\.[a-zA-Z]+$)/g;
+
+		let actualDomainMatch;
+		if (currentUrl) {
+			actualDomainMatch = currentUrl.hostname.match(regex);
+		} else if (typeof window !== undefined) {
```
gelbal: @laziob can you please update the check here?
```suggestion
		} else if (typeof window !== "undefined") {
```

[Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) says that `typeof` returns string.

gelbal (commented):
Ezequiel, I asked for one more change before we merge. Unless we hear from Amit or Stevche, I think it's fine to merge this PR. 

It's better to roll out this change asap and there is no strong reason to introduce further changes like balena domain tracking here.

src/url-params.ts
pranasziaukas: Wow, that's a great catch @gelbal, I completely missed (read: never knew) that!

laziob: Cant believe I missed that 🤦‍♂️  sorry for that! Thanks @gelbal !

gelbal: No worries. Thanks for the quick change.

gelbal:
@laziob thanks again for going through this PR dance. Nice work. Let's ship it.

Can you please squash your commits before we merge this?

laziob:
@gelbal there it goes! Thank you and @pranasziaukas for all the review so far! I promise I'll get better at coding eventually

pranasziaukas (approved):
Let's do it.

PR #48, opened 2021-03-01
Created new declareClient function un url-params

laziob:
This new function expects a a client to be passed, and sets this client
on the AnalyticsUrlParams class for it to be used on functions such as
getSessionId, getQueryString and so on.

Without it, those functions wont work with the new way to init
the analytics client.

Change-type: Minor

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ const client = createClient({
     componentVersion: require('./package.json').version, // (optional) automated version reporting
     deviceId: passedDeviceId,
 })
+
+// A newly created client is passed to the `AnalyticsUrlParams` to make it fully functional.
+urlParamsHandler.setClient(client);
 ```
 
 ## Usage
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -103,12 +103,24 @@ export class AnalyticsUrlParams {
 		if (this.client) {
 			throw new Error('Client is already set');
 		}
-		const newDeviceId = this.setDeviceIds(null, client.deviceId());
-		if (newDeviceId != null) {
-			client.setDeviceId(newDeviceId);
+		if (
+			this.getPassedDeviceId() &&
+			this.getPassedDeviceId() !== client.deviceId()
+		) {
+			const newDeviceId = this.setDeviceIds(null, client.deviceId());
+			if (newDeviceId != null) {
+				client.setDeviceId(newDeviceId);
+			}
+		}
+		if (this.sessionId && client.sessionId() !== this.sessionId) {
+			client.setSessionId(this.sessionId);
 		}
 		this.sessionId = client.sessionId();
 		this.client = client;
+	}
+
+	getClient() {
+		return this.client;
 	}
 
 	/**
--- a/test/url-params.test.ts
+++ b/test/url-params.test.ts
@@ -153,7 +153,10 @@ const clientMock = () =>
 
 		deviceId() {
 			this.deviceIdRetrieved = true;
-			return 'test_device_id';
+			if (!this.setDeviceIdParams) {
+				return 'test_device_id';
+			}
+			return this.setDeviceIdParams;
 		},
 		sessionId() {
 			return this.knownSessionId;
@@ -262,9 +265,45 @@ test('set default client', () => {
 	urlParams.consumeUrlParameters('optOutAnalytics=false');
 	expect(urlParams.isOptOutRequested()).toBeFalsy();
 
+	const passedDeviceId = urlParams.getPassedDeviceId();
 	const client = clientMock();
-	const initialDeviceId = client.deviceId();
+
+	if (passedDeviceId) {
+		client.setDeviceId(passedDeviceId);
+	}
 	urlParams.setClient(client);
-	expect(client.deviceIdRetrieved).toBeTruthy();
-	expect(client.setDeviceIdParams).toStrictEqual(initialDeviceId);
+	const allDeviceIds = urlParams.allDeviceIds();
+	expect(client.setDeviceIdParams).toStrictEqual(null);
+	expect(allDeviceIds).toEqual(['test_device_id']);
+	expect(client.knownSessionId).toStrictEqual(123);
+});
+
+test('set default client with passed sessionId', () => {
+	const urlParams = new AnalyticsUrlParams();
+	urlParams.consumeUrlParameters('s_id=999&optOutAnalytics=false');
+	expect(urlParams.isOptOutRequested()).toBeFalsy();
+	const passedSessonId = urlParams.getSessionId();
+
+	const client = clientMock();
+	urlParams.setClient(client);
+	expect(client.setDeviceIdParams).toStrictEqual(null);
+	expect(client.knownSessionId).toStrictEqual(passedSessonId);
+});
+
+test('set default client with passed deviceId', () => {
+	const urlParams = new AnalyticsUrlParams();
+	urlParams.consumeUrlParameters('d_id=999,888,777');
+
+	const passedDeviceId = urlParams.getPassedDeviceId();
+	const client = clientMock();
+
+	if (passedDeviceId) {
+		client.setDeviceId(passedDeviceId);
+	}
+	const allDeviceIds = urlParams.allDeviceIds();
+	urlParams.setClient(client);
+	expect(client.setDeviceIdParams).toStrictEqual('999');
+	expect(allDeviceIds).toEqual(['999', '888', '777']);
+	expect(client.knownSessionId).toStrictEqual(123);
+	expect(urlParams.getClient()).toEqual(client);
 });


README.md
```diff
@@ -40,6 +40,9 @@ const client = createClient({
     componentVersion: require('./package.json').version, // (optional) automated version reporting
     deviceId: passedDeviceId,
 })
+
+//And at last, we pass this newly created client to the AnalyticsUrlParams since its needed for some functionalities
```
pranasziaukas:
```suggestion
// A newly created client is passed to the `AnalyticsUrlParams` to make it fully functional.
```
No need to use this specific suggestion, but my idea is to unify the spacing, punctuation and maybe reduce the number of unnecessary filler words in the comments of the example snippet.

src/url-params.ts
```diff
@@ -111,6 +111,15 @@ export class AnalyticsUrlParams {
 		this.client = client;
 	}
 
+	/**
+	 *
+	 * @param client
+	 * Sets the class client value to whatever client object is passed on the argument.
+	 */
+	declareClient(client: Client) {
```
pranasziaukas: I think the idea is decent and it is a needed change.
Thinking further, do we have any other naming suggestions? I'm trying to imagine how confusing it is to have `setClient` and `declareClient`.

README.md
laziob: Yeap, it sounds better thay way!

src/url-params.ts
laziob: Right, I have the same feeling that is not the best name, but then again, what we are doing is actually setting the client value on the AnalyticsUrlParams class, but setClient is already taken for a much bigger purpose, and I thought that event having something like setClassClient , including both 'set' and 'client' would be more confusing than 'declareClient'. But Im open to any suggestion!

gelbal (commented):
Taking a step back, I'd ask us to start with: What's the exact friction point this PR wants to address?

Ezequiel explained that he doesn't like to re-do the work done on the initialization. Is that about not calling `setDeviceIds` for 2nd time?
(which writes the `passedDeviceId` to cookie)

Then how about setting `sessionId`? 
Right now we don't expose any way for the component to directly call `setSessionId` (as it's a private method). So using `declareClient`, the component doesn't have a way to set `sessionId`.

Instead of introducing such a new method, could we extend the implementation of `setClient` with extra parameter that prevents re-writing cookie? 
(assuming that's the only friction point)

laziob:
@gelbal  Right, I wanted to avoid rewriting cookies and going through all that code again since there was no purpose for it. When possible please take a look at the new approach for setClient(). I left come of the code commented with my observations as to why, but basically, there was some redundancy on its tasks due to how we handle deviceIds since the last change and I couldnt think of a case where that was needed now.

src/url-params.ts
```diff
@@ -103,12 +103,21 @@ export class AnalyticsUrlParams {
 		if (this.client) {
 			throw new Error('Client is already set');
 		}
-		const newDeviceId = this.setDeviceIds(null, client.deviceId());
-		if (newDeviceId != null) {
-			client.setDeviceId(newDeviceId);
-		}
-		this.sessionId = client.sessionId();
```
pranasziaukas: I can see how this logic is not straightforward to follow.

src/url-params.ts
```diff
@@ -103,12 +103,21 @@ export class AnalyticsUrlParams {
 		if (this.client) {
 			throw new Error('Client is already set');
 		}
-		const newDeviceId = this.setDeviceIds(null, client.deviceId());
```
pranasziaukas, line 106: `newDeviceId` can only be `null` if neither `client` nor the instance of `AnalyticsUrlParams` is aware of any device IDs whatsoever.
Otherwise, `this.setDeviceIds` effectively just sets the cookie.

src/url-params.ts
```diff
@@ -103,12 +103,21 @@ export class AnalyticsUrlParams {
...
-		if (newDeviceId != null) {
-			client.setDeviceId(newDeviceId);
```
pranasziaukas, line 108: `client.setDeviceId(newDeviceId)` is redundant because it will always set the same value. Good catch.

src/url-params.ts
pranasziaukas, line 110: So, as I understand, Ezequiel is saying that this is safe to remove because of our updated workflow/interaction with the `analytics-client`, right?

src/url-params.ts
```diff
@@ -103,12 +103,21 @@ export class AnalyticsUrlParams {
...
+			this.client.setSessionId(this.sessionId);
+			return;
+		}
+
+		// This can only return null if the function is called with no arguments, which I dont know if it makes sense to do that ever
+		// const newDeviceId = this.setDeviceIds(null, client.deviceId());
+		// In consecuence, this will always run and override the client deviceId with its already set same value.
+		// this is due to the fact that now client.deviceId() has already been set to the one of the inputIdString.
+		// if (newDeviceId != null) {
+		// 	client.setDeviceId(newDeviceId);
```
pranasziaukas: I think this could've been a PR comment instead of the in-code comment. Or is it on purpose?

README.md
```diff
@@ -40,6 +40,9 @@ const client = createClient({
     componentVersion: require('./package.json').version, // (optional) automated version reporting
     deviceId: passedDeviceId,
 })
+
+// A newly created client is passed to the `AnalyticsUrlParams` to make it fully functional.
+urlParamsHandler.declareClient(client);
```
pranasziaukas: I believe this is supposed to be `setClient`.

pranasziaukas (commented):
Looks promising. @laziob just please review the unit tests to match the adjusted functionality.

src/url-params.ts
laziob: Exactly. On the old workflow, we were first creating a client, and then overriding that client `deviceId` with the first deviceId that was passed on the url, if any through `setDeviceIds()`. But if there was no `passedDeviceId` then `consumeUrlParams()` never called `setDeviceIds` , it just left the client with its own deviceId. So if then we pass `setClient()` that same client that already exists, we would be forcing the function just to return the same value. The only thing that now that I've written it makes me doubt about this change is that if we never call `setDeviceIds()` either via `consumeUrlParams()` or via `setClient()` (for example when there is no `passedDeviceId`) then we are never writting the current deviceId on the "__analytics_dids" cookie and maybe we dont want that.

laziob: But giving it some more thought, if there was no `psasedDeviceId` then that cookie was only written if the component maintainer called `setClient()` at any time for some reason. Maybe to know if we should leave the `setDeviceIds()` call or not, we should first understand in which cases the maintainer would have called `setClient()` before this changes where there was no `passedDeviceId`. Im going to look at the components code to see if I find any case,

laziob: Indeed it seems that we had some unwanted behaviour with how it was written. Look at [this example](https://github.com/balena-io/docs/blob/3183fdc0f98ea21487b86635dc0f693183461fa2/templates/_track.html#L29) on Docs. 
First it parses the url query string. So from here we have two cases, A and B.
A.1) If there is a `passedDeviceId`, then `consumeUrlParams()` would call `setDeviceIds()` with only the query string of passed deviceIds, since at that moment `originalDeviceId` would have returned `null` because there was no client set on AnalyticsUrlParams yet. As a result it writes the `__analytics_dids` cookie with only the string of `passedDeviceIds` and then it returns the first one from that list.
A.2) After that, the client is created and then it calls `setClient` with said client. So `setClient()` calls `setDeviceIds` again, only this time it does not passes any `inputIdString` on the first argument, but it does pass a `client.deviceId()` on the second one. As a result the cookie gets updated due to the concat and appends the client deviceId to it, but since it returns the `client.deviceId()` as the `newDeviceId` of `setClient()`, it then overrides its already set `deviceId` with it same value. and ay `passedDeviceId` only gets saved on the cookie but does not get used. 

B.1) If there is no `passedDeviceId` then `consumeUrlParams()` would not call `setDeviceIds()`. As a result, we dont have any "__analytics_dids" cookie written yet nor any list of deviceIds saved.
B.2) Then after client is created, it calls `setClient` and it passes the deviceId of said client to `setDeviceIds()`. As a result we now have the cookie with only the deviceId created by the client itself, and then we override that value with its same value.

So what we need is a way for the cookie to be written with all the deviceIds, best if we dont have to write it 2 times on a same initialization, also to make sure that we are using a `passedDeviceId` for the client (this I think its already covered) and last not override the client deviceId with its same value. 

I'll push some changes for a way we might be able do all this but please let me know of any suggestions.

src/url-params.ts
laziob: It was on purpose for it to appear directly here but I will remove it before we merge

README.md
laziob: Right thanks! changed it.

src/url-params.ts
gelbal: Thanks for digging into this and explaining your thought process @laziob. I agree that we could refactor the code here to make it cleaner and easier to read. I still believe we could achieve these while re-using the same function (`setClient`), so also thanks for pushing that idea further.

Shall we introduce a `getDeviceId()` method to `AnalyticsUrlParams` that returns the `deviceId` that it'd like to set on the client?
We could compare the output of this method with `client.deviceId()` before re-setting the same value. Or alternatively, we could build a new method to check if `client.deviceId()` is already part of `AnalyticsUrlParams.deviceIds` before we take any action. 

My reasoning is we already have 1 place where we modify `AnalyticsUrlParams.deviceIds` and set the relevant cookie. Let's keep it like that. The main thing missing is a check that prevents double work here (and potential bugs).

laziob: @gelbal I thought about it but the comparison doesnt make sense because if we use `client.deviceId()` for comparison then it means we are already late because we already init a client with a maybe wrong deviceId. Thats why most of the work I think should be done by `consumeUrlParams()`. This should return the one `deviceId` that we want to use, if any, which I think it now does. And then in case we are not passing any deviceId so consumeUrlParams() doesnt return anything to set, now `setClient()` should be able to see that and write the cookie with the client that the init created.

I cant think of a case where we would need a `getDeviceId()` that happens before init , unless we use it for reading cookie values for some reason. For after init we already have the `getQueryString()` which does that and more. 

Also, what do you think about the tests I wrote? Im asking because Im still not that confident of doing them properly lol

gelbal: @laziob, I'd like to keep `analytics-client` rather generic and modular. This implies not overly customizing the implementation to cater to our existing use cases only. Following this logic, it'd be great to keep enabling the scenario of using `AnalyticsUrlParams` without the `client` part. 

Additionally, we already have the use case of setting client to `NoopClient` as seen [here](https://github.com/balena-io/resin-site/blob/64f6b968253a1c4052eacf751929b0fe32f05b78/src/utils/analytics-browser.js#L24). So it's actually required for `setClient` to support overwriting (essentially deleting) existing `deviceId` values and cookies.

Before we get to start writing implementation or tests, it'd be great to clearly define the problem here. What's the friction point?

laziob: @gelbal as for my understanding the friction point is that we need to pass a client to `AnalyticsUrlParams` in order for `getQueryString` or any get function that takes values from a client work when called afterwards. At the same time, ideally we would like to set a client without having to override or repeat work that has already been done on client init.  

As per our current init order, we 
1. call AnalyticsUrlParams and consumeUrlParams
2. Init a client using, if theres any, the passed values from consumeUrlParams

But so far, AnalyticsUrlParams does not know of any initialized client. Thats where `setClient()` comes in.
Although the use case missing for `setClient()` is to be able to just pass a initialized client to the class. The problem rises when the rest of the setClient code runs, since it does more than what is needed at the moment, and it does thing that wouldnt be convenient to do after a client is already initialized, like overriding its cookies.

Maybe we could live it as it was on the first place, and figure a way to let it know that a client has already been set, and if so, to just pass the client to the class and thats it.

gelbal: > since it does more than what is needed at the moment

I hear you @laziob. I'd like to get to the bottom of this redundant work. Could we prevent it with a refactor like this?
```js
setClient(client: Client) {
	if (this.client) {
		throw new Error('Client is already set');
	}
	if (this.getPassedDeviceId() && this.getPassedDeviceId() !== client.deviceId()) {
		const newDeviceId = this.setDeviceIds(null, client.deviceId());
		if (newDeviceId != null) {
			client.setDeviceId(newDeviceId);
		}		
	}
	
	this.sessionId = client.sessionId();
	this.client = client;
}
```

So we don't touch `deviceId` unless there is a discrepancy.

Next, we might consider initializing `Client` with `AnalyticsUrlParams` as part of the `config` object in the constructor. So we eliminate the extra line for `AnalyticsUrlParams.setClient(client)`.

laziob: Looks good! Let me test it and if everything is ok I'll commit it.

Then, I also think it would be better if the `client` is set within `AnalyticsUrlParams` directly from the client Init, either as part of config or as part of `createClient()` but I would have to look more in depth to that code first.

laziob: @gelbal were you able to look at the last commit that adresses this suggestion?

pranasziaukas (commented):
IMO this looks good. Just giving others the chance to check before merging.

src/url-params.ts
```diff
@@ -103,9 +103,18 @@ export class AnalyticsUrlParams {
...
+			this.getPassedDeviceId() !== client.deviceId()
+		) {
+			const newDeviceId = this.setDeviceIds(null, client.deviceId());
+			if (newDeviceId != null) {
+				client.setDeviceId(newDeviceId);
+			}
+		}
+		if (this.sessionId && client.sessionId() !== this.sessionId) {
+			client.setSessionId(this.sessionId);
+			return;
```
gelbal: Why does the code `return` here? Any particular reason to skip the rest of this function?

I'd expect this to be `if .. else` statement instead and we call `this.client = client;` for both flows.

test/url-params.test.ts
```diff
@@ -262,9 +265,48 @@ test('set default client', () => {
...
 	expect(urlParams.isOptOutRequested()).toBeFalsy();
 
+	const passedDeviceId = urlParams.getPassedDeviceId();
 	const client = clientMock();
-	const initialDeviceId = client.deviceId();
+
+	if (passedDeviceId) {
+		client.setDeviceId(passedDeviceId);
+	} else {
+		client.deviceId();
```
gelbal: I believe this `else` statement is redundant here as well as the following test below.

test/url-params.test.ts
```diff
@@ -262,9 +265,48 @@ test('set default client', () => {
...
+	const urlParams = new AnalyticsUrlParams();
+	urlParams.consumeUrlParameters('s_id=999&optOutAnalytics=false');
+	expect(urlParams.isOptOutRequested()).toBeFalsy();
+	const passedSessonId = urlParams.getSessionId();
+
+	const client = clientMock();
+	urlParams.setClient(client);
+	expect(client.setDeviceIdParams).toStrictEqual(null);
+	expect(client.knownSessionId).toStrictEqual(passedSessonId);
+});
```
gelbal, line 291: Shall we add one more check here to verify that `AnalyticsUrlParams.client` is equal to the `clientMock()`?

If we had that check, the test would had failed because of my earlier comment on the `return` part after sessionId setting :nerd_face:

gelbal (commented):
Thanks for the changes Ezequiel. I left couple of comments to address before we deploy. Also can you please squash your commits?

src/url-params.ts
laziob: Yeap that was an error on my side. I just fixed it. Thanks for noticing!

test/url-params.test.ts
laziob: Right, I see now how it was redundant since `setClient` should handle that work now.

test/url-params.test.ts
laziob: @gelbal  How can we do that? `AnalyticsUrlParams.client` doesnt exists per se and we dont have a method yet to retrieve the client. Even more, the variable names on each are also different, like for example `knownSessionId`

src/url-params.ts
gelbal: Cool. @laziob, I don't see any new commit on this repo since my review.
(since you say "I just fixed it")

test/url-params.test.ts
gelbal: Right. We could implement `getClient()` method and compare like:
```javascript
expect(urlParams.getClient()).toEqual(client);
```

Not sure if there is any other use case for `getClient()` honestly. Then it would have caught that little bug with client setting.

src/url-params.ts
laziob: I have it on local, was going to push once I get the other thing on the test.

test/url-params.test.ts
laziob: Ah ok. Yeah I was asking like, if it is worth implementing a getClient method just for that test. And also, I dont understand that ClientMock really well so I dont know if I should expect it to be equal to urlParams Client since they have different variables names and stuff. But if you say it should, then I'll go on and implement getClient and see what happens. let me try and push it so you can see.

laziob:
For some reason, after I merge master into this branch, squash that merge commit to the older one and push, it asks me again to merge master into this branch :S . Aside from that I think the rest of the commits were all properly squashed.

laziob:
@balena-ci rebase

gelbal (approved):
Let's ship it.

PR #49, opened 2021-04-22
Added unsetParamsReferrerOnNewSession to client

laziob:
With this config variable set to true, now new sessions that
dont include referrer or UTM values will be set to null or none
instead of carrying over previous referrers. This should help us
to properly identify trends and behaviour of referrers since
its more transparent than the previous behaviour.

Change-type: Minor

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -97,6 +97,7 @@ class DefaultClient implements Client {
 		amplConfig.includeReferrer = true;
 		amplConfig.includeUtm = true;
 		amplConfig.sameSiteCookie = 'Lax';
+		amplConfig.unsetParamsReferrerOnNewSession = true;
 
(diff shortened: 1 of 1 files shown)


pranasziaukas (approved):
Sounds good Ezequiel.

You explained it already, but I peeked at the docs anyway :)
> By default, the SDK will carry over existing UTM Parameters and Referrer values at the start of a new session. For example, if a users session expires, the SDK will map the user's Referrer and UTM Parameters to existing values. To reset those values to null upon instantiating a new session, set `unsetParamsReferrerOnNewSession` to `true`.

https://developers.amplitude.com/docs/advanced-settings#tracking-utm-parameters-referrer-and-gclid-javascript

PR #50, opened 2021-05-12
Get getQueryString to expect relative URLs too

laziob:
Change-type: Minor

### DIFF
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -189,8 +189,8 @@ export class AnalyticsUrlParams {
 			try {
 				destinationUrl = new URL(destinationUrl);
 			} catch (err) {
-				console.error(err)
-				return ''; 
+				console.error(err);
+				return '';
 			}
 		}
 
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -189,7 +189,8 @@ export class AnalyticsUrlParams {
 			try {
 				destinationUrl = new URL(destinationUrl);
 			} catch (err) {
-				return ''; //I dont know what we should do when an error like this happens. Any ideas??
+				console.error(err)
+				return ''; 
 			}
 		}
 
--- a/src/url-params.ts
+++ b/src/url-params.ts
@@ -175,12 +175,34 @@ export class AnalyticsUrlParams {
 	/**
 	 * @return full query parameter string that can be appended to URLs
 	 */
-	getQueryString(destinationUrl?: URL, currentUrl?: URL): string {
+	getQueryString(destinationUrl?: URL | string, currentUrl?: URL): string {
+		// we first check if destionationUrl is a relative URL string. If it is, we exit.
+		const relativeRegex = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/g;
+		if (
+			typeof destinationUrl === 'string' &&
+			destinationUrl.match(relativeRegex)
+		) {
+			return '';
+		}
+
+		if (typeof destinationUrl === 'string') {
+			try {
+				destinationUrl = new URL(destinationUrl);
+			} catch (err) {
+				return ''; //I dont know what we should do when an error like this happens. Any ideas??
+			}
+		}
+
 		// this regex is based on the assumption that we wont be using TLDs longer than 3 characters. If we do, it will break
 		// the logic and take that longer TLD as the main domain, for example hub.balena.edge.io -> edge.io
 		const regex = /([a-zA-Z0-9-]+)(\.[a-zA-Z]{2,3})?(\.[a-zA-Z]+$)/g;
 
-		let actualDomainMatch;
+		const destinationDomainMatch = destinationUrl
+			? destinationUrl.hostname.match(regex)
+			: null;
+		const destinationDomain = destinationDomainMatch?.[0];
+
+		let actualDomainMatch: RegExpMatchArray | null;
 		if (currentUrl) {
 			actualDomainMatch = currentUrl.hostname.match(regex);
 		} else if (typeof window !== 'undefined') {
@@ -189,16 +211,7 @@ export class AnalyticsUrlParams {
 			actualDomainMatch = null;
 		}
 
-		const destinationDomainMatch = destinationUrl
-			? destinationUrl.hostname.match(regex)
-			: null;
-
-		const actualDomain = actualDomainMatch
-			? actualDomainMatch.toString()
-			: null;
-		const destinationDomain = destinationDomainMatch
-			? destinationDomainMatch.toString()
-			: null;
+		const actualDomain = actualDomainMatch?.[0];
 
 		if (!destinationDomain || actualDomain !== destinationDomain) {
 			return [this.getDeviceIdsQueryString(), this.getSessionIdQueryString()]
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@types/js-cookie": "^2.2.5",
     "@types/lodash": "^4.14.142",
     "@types/mixpanel-browser": "2.35.1",
+    "@types/node": "^16.6.1",
     "husky": "^3.0.9",
     "jest": "^24.9.0",
     "lint-staged": "^9.4.2",
--- a/test/url-params.test.ts
+++ b/test/url-params.test.ts
@@ -137,6 +137,27 @@ test('parsing and matching destination and actual URL to regex', () => {
 			new URL('https://domain2.edge.io'),
 		),
 	).toBe('');
+
+	// Case when passing a relative URL as destinationUrl
+	expect(
+		urlParams.getQueryString('/etcher', new URL('https://domain2.edge.io')),
+	).toBe('');
+
+	// Case when passing an absolute URL as a string for destinationUrl and no passing is expected
+	expect(
+		urlParams.getQueryString(
+			'https://test.domain.io',
+			new URL('https://domain.io'),
+		),
+	).toBe('');
+
+	// Case when passing an absolute URL as a string for destinationUrl and passing is expected
+	expect(
+		urlParams.getQueryString(
+			'https://test.domain.io',
+			new URL('https://otherdomain.com'),
+		),
+	).toBe('d_id=d1&s_id=123');
 });
 
 interface AnalyticsMock {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,8 @@
 {
   "compilerOptions": {
+    "lib": [
+      "ES6","DOM"
+    ],
     "module": "commonjs",
     "target": "es5",
     "outDir": "dist",


laziob:
@gelbal @pranasziaukas I added the logic to handle both relative and absolute Urls and corresponding tests that seem to be working. But as always, please give it a look since you might come up with a better way to handle this or notice something I missed.

Thanks!

gelbal:
@laziob I see balenaCI returns failed because of not-prettified code. Can you please run `npm run prettify` and then commit again to clear these errors?

> Error: File src/url-params.ts hasn't been formatted with prettier
Error: File test/url-params.test.ts hasn't been formatted with prettier

I wonder if it makes sense to include `prettify` in `lint` to catch these issues earlier :thinking: 
(I'm reviewing the rest of PR now)

src/url-params.ts
```diff
@@ -175,11 +175,24 @@ export class AnalyticsUrlParams {
...
+		const relativeRegex = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/g;
+		if (
+			typeof destinationUrl === 'string' &&
+			destinationUrl.match(relativeRegex)
+		) {
+			return '';
+		}
+
+		// If its not a relative URL string, we make sure if its a string or an URL. It its a string, we convert it to an URL object.
+		if (typeof destinationUrl === 'string') {
```
gelbal, line 188: You can combine this `if` statement with the one above to eliminate the double check for `string`. The code and comments will also read better.
```js
if (typeof destinationUrl === 'string') {
    if (destinationUrl.match(relativeRegex)) {
        return '';
    } else {
        ....
}

File: src/url-params.ts [hunk 2]
```diff
@@ -175,11 +175,24 @@ export class AnalyticsUrlParams {
 	/**
 	 * @return full query parameter string that can be appended to URLs
 	 */
-	getQueryString(destinationUrl?: URL, currentUrl?: URL): string {
+	getQueryString(destinationUrl?: URL | string, currentUrl?: URL): string {
+		// we first check if destionationUrl is a relative URL string. If it is, we exit.
+		const relativeRegex = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/g;
```
gelbal, line 180: So this regex catches strings like `hub.balena.io` or `balena.io`. Do we want that? Shall we exclude the strings with domain names as well?

src/url-params.ts
```diff
@@ -175,11 +175,24 @@ export class AnalyticsUrlParams {
...
+		if (
+			typeof destinationUrl === 'string' &&
+			destinationUrl.match(relativeRegex)
+		) {
+			return '';
+		}
+
+		// If its not a relative URL string, we make sure if its a string or an URL. It its a string, we convert it to an URL object.
+		if (typeof destinationUrl === 'string') {
+			destinationUrl = new URL(destinationUrl);
```
gelbal: We need to handle the errors here when the incoming string cannot be translated to a URL. I'm not sure if `relativeRegex` ensures that the string with no match is actually a URL.

gelbal (commented):
Good stuff pushing this forward Ezequiel. I typed couple of suggestions and discussion points.

src/url-params.ts
laziob: @gelbal  the thing is that an href thats only 'hub.balena.io' or 'hub.balena.io' shoudnt be possible since it wont work. If you use that as an `href` the HTML would send you to www.domain.com/hub.balena.io . I think that should be always be avoided in practice right?

src/url-params.ts
laziob: @gelbal Ohh yeap you are right. let me look into this.

src/url-params.ts
laziob: @gelbal Lol I was going to do it that way first, but for some reason I remembered that once I saw the if statements one below the other and that that was the usual practice or something like that and went the other way. Sure I'll change that.

gelbal: @laziob oh I see you made a major rework. Actually I liked your previous implementation as it was cleaner (once you combine the new `if` statements). So the code handles the `string` scenario first and the proceeds to handling the `URL` scenario. Can you please bring it back? : )

gelbal: Petros (ex-CTO) made a pretty good presentation about such design principle actually (he calls it `normalizing if`). I highly recommend listening to it: 
https://docs.google.com/presentation/d/1hFV5YBL4FflfN7OIJr6eZrUV4Pk158wOXWzea1ANTnM/edit#slide=id.p
https://drive.google.com/file/d/1xtHev7HUwbASdasf8cWfG9O2oHdlXX43/view

src/url-params.ts
gelbal: Aha. Good to know. Then this regex seems like working fine.

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
...
+				destinationUrl = new URL(destinationUrl);
+				destinationDomainMatch = destinationUrl
+					? destinationUrl.hostname.match(regex)
+					: null;
+
+				destinationDomain = destinationDomainMatch
+					? destinationDomainMatch.toString()
+					: null;
+			} catch (err) {
+				return ''; //I dont know what we should do when an error like this happens. Any ideas??
```
gelbal: Good question. We could do `console.log(err)` but I don't know how much that'd help. I think it's fine to at least catch it as you do here.

src/url-params.ts
laziob: @gelbal It still handles the string scenario first. The code above is for the other parameter that stays that way regardless of the `destinationUrl`. The reason I changed it was because with the previous design, I had to repeat much more code, once for each if and even then couldnt make it work properly. In fact I dont even like this design, since it still repeats code. But it repeated less than the previous one. I still also dont believe we should be handling relative URLs through this method. It kind of undermines the use of TypeScript and gives place to the bad use of the method in `resin-site`, since I still think they shoudnt be forcing this method in every link "just in case". Performance wise it doesnt make sense to me. `resin-site` should have a better way to identify outbound links from inbound links, and call this method only on outbound links. 

Having said that, I could try go back to the previous design, but I could make it work with the previous design. This was so far the only way I achieved what I wanted lol.

gelbal: > It kind of undermines the use of TypeScript

I know what you mean. I checked and apparently TypeScript does not yet support string types based on regex. See the issue: https://github.com/microsoft/TypeScript/issues/41160
Regex-validated string types would be great but alas we don't have them now.

As for `normalizing if`: Your previous solution was cleaner as it first handles the scenario of receiving `string` parameter and then once it's handled, the rest of the logic stays the same for `URL` parameter scenario.

gelbal: Is there a simpler way for UI / React developers to signal relative URLs on their end? 

I understand the frontend folks' point of view as they want a convenient method that handles all scenarios to generate URL params. I don't think it's bad or lazy on their side to pass any URLs to this function. At the end `analytics-client` is a module and you could think of it as an extension of the tools that frontend folks use as part of their own codebase. So it's totally fine for the module to be responsible for such validations.

laziob: By no means Im saying that its because of lazyness or unwillingess to do it. Is just that adding the URL params is something that has to be done in really few cases relative to the amount of links a web component might have. So calling if for all links seems backwards. I dont know much about performance to be honest, but is like building a kind of sorter, when actually the cases that you need to find positive are really few and can be pinpointed. For example the way it is now, we would also be appending Url Params to even external links like twitter.com, github.com, youtube.com, and many sites that dont know what does params even are, which it might even break down links since we dont control what happens on the other side when URL params are present. So to solve for that this method should also handle a way to identify just balena domains. For that we would then need another regex (which are not bulletproof) or a list somewhere that would need maintenance. It feels like building a big machine for doing a small task.

gelbal: I hear you Ezequiel. I agree that it'd be neater if the module user has a simple way to detect relative URLs (and also external 3rd party sites) on their app code, and in such cases not call this `getQueryString` method . Doesn't seem like so, right? 

Yeah, an explicit list of domains to keep tracking would be neat. I know we discussed this briefly before. Perhaps we could revisit this topic, given the increasing complexity when we don't have such an explicit list. It might not be too much to ask from component maintainers to provide `analytics-client` a list of potential domains that a user could navigate to.

laziob: To put an example, `Hub` handles it like this:

`	deployUrl = getExternalLinkWithTracking(
		'https://dashboard.balena-cloud.com/deploy?repoUrl=${repoUrl}')`

So it explicitly uses the method on the only domain where its needed. For example marketing site only needs to use it on the Login and Signup buttons, which exist only in 2 places, the header (for most of the resin-site pages) , like [here](https://github.com/balena-io/resin-site/blob/90a8eaf39cc33298e558a744171b8ccd58e38d5c/src/components/Header/Header.js#L246) and the box for on the landing page. 

So its not that they need to detect relative urls in a simple way. Its that they have to detect URLs to another balena domain, not even including subdomains. 

If using a list is the best way to do it I dont really know, but if its needed to be raised for further discussion Im up for that. Whats the proper way and place to raise it?  ALthough It just seems too much for the task at hand tbh.

laziob: @gelbal we need to work on this soon since its the only thing preventing marketing sites to update to the last analytics client. Its still using 0.12 so its not stiching sessions nor using the last referrer updates.

laziob: @iamsolankiamit @thgreasi @JSReds if you could take a look into the discussion too so we can decide on how we advance it would be really helpful and appreciated! thanks!

src/url-params.ts
```diff
@@ -175,12 +175,18 @@ export class AnalyticsUrlParams {
...
+		let destinationDomainMatch;
+		let destinationDomain;
+		let actualDomainMatch;
+		let actualDomain;
+
 		// this regex is based on the assumption that we wont be using TLDs longer than 3 characters. If we do, it will break
 		// the logic and take that longer TLD as the main domain, for example hub.balena.edge.io -> edge.io
 		const regex = /([a-zA-Z0-9-]+)(\.[a-zA-Z]{2,3})?(\.[a-zA-Z]+$)/g;
 
-		let actualDomainMatch;
```
thgreasi, line 183: I actually prefer it here, next to where it's first assigned, so that's easier to follow.
Also type it as:
```
let actualDomainMatch: RegExpMatchArray | null;
```

src/url-params.ts
```diff
@@ -175,12 +175,18 @@ export class AnalyticsUrlParams {
...
 	 * @return full query parameter string that can be appended to URLs
 	 */
-	getQueryString(destinationUrl?: URL, currentUrl?: URL): string {
+	getQueryString(destinationUrl?: URL | string, currentUrl?: URL): string {
+		// we first check if destionationUrl is a relative URL string. If it is, we exit.
+		//const relativeRegex = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/g;
+		let destinationDomainMatch;
+		let destinationDomain;
+		let actualDomainMatch;
+		let actualDomain;
```
thgreasi: All these are typed as `any`, which isn't good...
After the changes suggested bellow this will be:
```suggestion
		let destinationDomain: string | undefined;
```
but I would suggest to push it lower, near the place it's first assinged.

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
 			actualDomainMatch = null;
 		}
 
-		const destinationDomainMatch = destinationUrl
-			? destinationUrl.hostname.match(regex)
-			: null;
+		actualDomain = actualDomainMatch ? actualDomainMatch.toString() : null;
```
thgreasi:
```suggestion
		const actualDomain = actualDomainMatch ? actualDomainMatch.toString() : null;
```
and if you have recent enough TS in this project, then you can do:
```
		const actualDomain = actualDomainMatch?.[0]; // this is inferred as `string | null` automatically
```
prefer explicitly reading the first match, that relying on `toString()`.

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
...
+		if (typeof destinationUrl === 'string') {
+			try {
+				destinationUrl = new URL(destinationUrl);
```
thgreasi, line 201: Do you really need to pass this through the URL constructor again?
If there is a chance of this not being a URL (string), then this should be in its declaration.

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
...
+				destinationDomainMatch = destinationUrl
```
thgreasi:
```suggestion
				const destinationDomainMatch = destinationUrl
```
Just declare it in this scope, since this is only used in this limited scope as a temp variable and its value isn't useful outside this context.

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
...
+					: null;
+
+				destinationDomain = destinationDomainMatch
+					? destinationDomainMatch.toString()
+					: null;
+			} catch (err) {
+				return ''; //I dont know what we should do when an error like this happens. Any ideas??
+			}
+		} else {
+			destinationDomainMatch = destinationUrl
```
thgreasi:
```suggestion
			const destinationDomainMatch = destinationUrl
```
> Just declare it in this scope, since this is only used in this limited scope as a temp variable and its value isn't useful outside this context.

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
...
+		if (typeof destinationUrl === 'string') {
+			try {
+				destinationUrl = new URL(destinationUrl);
+				destinationDomainMatch = destinationUrl
+					? destinationUrl.hostname.match(regex)
+					: null;
+
+				destinationDomain = destinationDomainMatch
+					? destinationDomainMatch.toString()
+					: null;
```
thgreasi:
```suggestion
				destinationDomain = destinationDomainMatch?.[0];
```

src/url-params.ts
```diff
@@ -189,16 +195,29 @@ export class AnalyticsUrlParams {
...
 
-		const actualDomain = actualDomainMatch
-			? actualDomainMatch.toString()
-			: null;
-		const destinationDomain = destinationDomainMatch
-			? destinationDomainMatch.toString()
-			: null;
+			destinationDomain = destinationDomainMatch
+				? destinationDomainMatch.toString()
+				: null;
```
thgreasi:
```suggestion
			destinationDomain = destinationDomainMatch?.[0];
```

src/url-params.ts
gelbal: Hey @laziob cc @thgreasi @JSReds @iamsolankiamit, how about introducing an optional parameter (in the `analytics-client` constructor) for a list of domains to track?

We could retrieve this list either directly from the client that calls `analytics-client` or indirectly from the `projectName` configuration stored in the `analytics-backend`. At first it'll have the following content:
```
*.balena.io
*.balena-cloud.com
```
We will probably end up updating this list seldom so it's not a big maintenance issue to store it on the `analytics-backend` side.

---

Edit: Thinking it further, part of my suggestion actually means that `analytics-client` needs to retrieve such stateful information from the `analytics-backend` at each initialization. So maybe that's a no-go. It's cleaner for the component to configure `analytics-client` directly (without dependency on the backend).

gelbal: Having said these, let's still go ahead with working on this PR to add support for relative URL detection.

laziob:
@gelbal I think I properly reverted to the previous way we handled the relative URLs, hadnling string first and so but keeping some error handling. The unit tests are still working properly too. So please take a look when possible. On the other hand, I still dont know whats the best way to pass href and window.location.hostname as arguments to getQueryString() on the resin-site side [here](https://github.com/balena-io/resin-site/blob/8a20acdf189d869ad5620913ed5396fe09c38b7a/src/components/Link.js#L99) which is adressed on the PR though but with no solution yet

src/url-params.ts
```diff
@@ -175,30 +175,42 @@ export class AnalyticsUrlParams {
...
+		// we first check if destionationUrl is a relative URL string. If it is, we exit.
+		const relativeRegex = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/).*/g;
+		if (
+			typeof destinationUrl === 'string' &&
+			destinationUrl.match(relativeRegex)
+		) {
+			return '';
+		}
+	
+		// If its not a relative URL string, we make sure if its a string or an URL. It its a string, we convert it to an URL object.
```
gelbal: I don't think we need a comment here as the code is self explanatory.

src/url-params.ts
gelbal: @thgreasi @iamsolankiamit any suggestions on handling such caught errors in modules? 

It's OK to silently fail here but I wonder what's the best practice.

gelbal (commented):
Good stuff @laziob. I like this version the best. 

The error you get is:
`Error: File src/url-params.ts hasn't been formatted with prettier`

Can you please call `npm run prettify` and squash all your commits before pushing once more?

gelbal:
Hey @laziob, clicking on the `Details` button next to the checks takes you to the logs of balena CI. We see the following lint error:
```
src/url-params.ts:192:18
ERROR: 192:18  comment-format  comment must start with a space
```

It's best to get rid of that comment anyhow. Can you please add `console.error(err)` before returning? 

Exposing the error seems more useful than harmful. 

On other news, is this PR waiting for some other changes?

laziob:
@balena-ci I self certify!

PR #54, opened 2022-09-07
rolls back analytics client version

otaviojacobi:
Change-type: patch
Signed-off-by: Otavio Jacobi <otavio.jacobi@balena.io>


otaviojacobi:
@balena-ci I self certify!

PR #55, opened 2022-09-26
Replace balenaCI with flowzone

otaviojacobi:
Change-type: patch

### DIFF
--- a/repo.yml
+++ b/repo.yml
@@ -1,0 +1,1 @@
+type: 'node'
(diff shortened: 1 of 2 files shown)


repo.yml
```diff
@@ -0,0 +1 @@
+type: 'node'
```
Page-: You no longer need this file and can remove it 🎉

otaviojacobi: Thanks!

otaviojacobi:
@balena-ci I self-certify!

PR #60, opened 2025-04-15
Guides and surveys

myarmolinsky:
Improvement: https://balena.fibery.io/Work/Improvement/Add-the-Amplitude-Guides-and-Surveys-SDK-to-the-dashboard-2729
Project: https://balena.fibery.io/Work/Project/Enabling-experimentation-at-Balena-1108
Zulip: [#balena-io/balenaCloud > Add Amplitude Guides And Surveys SDK in the UI @ 💬](https://balena.zulipchat.com/#narrow/channel/346007-balena-io.2FbalenaCloud/topic/Add.20Amplitude.20Guides.20And.20Surveys.20SDK.20in.20the.20UI/near/512552309)
Amplitude docs: https://amplitude.com/docs/guides-and-surveys/sdk


src/client.ts
```diff
@@ -89,6 +90,7 @@ class DefaultClient implements Client {
 
 	constructor(config: Config) {
 		this.amplitudeInstance = createInstance();
+		this.amplitudeInstance.add(engagementPlugin());
```
JSReds, line 93: do we want to add it for the marketingClient as well ? https://github.com/balena-io-modules/analytics-client/blob/e125e54e758b1778190ca6b832acb41a33e6b42e/src/marketingClient.ts#L27

JSReds (commented):
Can we just add a small description with maybe a doc lik to explain what we are adding please ? 

Other then that LGTM 🎉

src/client.ts
myarmolinsky: Not at the moment

JSReds (approved):

PR #64, opened 2025-09-03
Update Amplitude engagement plugin

drskullster:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "dependencies": {
     "@amplitude/analytics-browser": "^2.16.0",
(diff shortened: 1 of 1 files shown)


myarmolinsky (approved):

src/client.ts
```diff
@@ -91,6 +91,7 @@ class DefaultClient implements Client {
 
 	constructor(config: Config) {
 		this.amplitudeInstance = createInstance();
+
```
myarmolinsky: nit: remove this line and file change? 😅

PR #67, opened 2026-03-25
Add amplitude instance to the global scope

drskullster:
Change-type: minor

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Convenient builders to compose analytics tools",
   "repository": {
     "type": "git",
-    "url": "https://github.com/balena-io/analytics-client"
+    "url": "git+https://github.com/balena-io-modules/analytics-client.git"
   },
   "publishConfig": {
     "access": "public"
--- a/src/marketingClient.ts
+++ b/src/marketingClient.ts
@@ -51,6 +51,9 @@ class MarketingClient implements Client {
 		const identifyObject = getIdentifyObject();
 		identifyObject.set(USER_PROP_COMPONENT_NAME, config.componentName);
 		this.amplitudeInstance.identify(identifyObject);
+		
+		// @ts-expect-error overriding the global scope
+		window?.amplitude = this.amplitudeInstance;
 	}
 
 	deviceId(): string {


src/marketingClient.ts
```diff
@@ -10,6 +10,12 @@ import {
 	USER_PROP_ANALYTICS_CLIENT_VERSION,
 	USER_PROP_COMPONENT_NAME,
 } from './config';
+import { BrowserClient } from "@amplitude/analytics-core";
```
otaviojacobi: Tiny detail @drskullster but would you mind changing it to 
```suggestion
import type { BrowserClient } from "@amplitude/analytics-core";
```
as it is strictly more correct?

src/marketingClient.ts
```diff
@@ -10,6 +10,12 @@ import {
...
+
+// Interface for the window object
```
otaviojacobi: Not necessary comment really

src/marketingClient.ts
```diff
@@ -10,6 +10,12 @@ import {
...
+declare global {
+	interface Window { amplitude: BrowserClient; }
+}
```
otaviojacobi: Maybe we would declare this in a specific global.d.ts file instead of it being in a generic file in the lib? Like we do in the CLI, for example: https://github.com/balena-io/balena-cli/blob/04dfac508ab0bd41a486bbc6b7ce617bc80acd7b/typings/global.d.ts#L4

otaviojacobi: + one question are you sure this global/Window is accessible from within the UI? I am also concerned that this might mess up with the typings on other repos. Maybe, instead of adding a typing for it we just assign `window.amplitude = ` and ts-expect-error explaining it on the below assignment?

drskullster: @otaviojacobi if we use the amplitude browser sdk that means that we necessarily have a window object, otherwise things will fail badly. I updated the code to use `ts-expect-error`

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Convenient builders to compose analytics tools",
   "repository": {
     "type": "git",
-    "url": "https://github.com/balena-io/analytics-client"
+    "url": "https://github.com/balena-io-modules/analytics-client"
```
otaviojacobi:
```suggestion
    "url": "git+https://github.com/balena-io-modules/analytics-client.git"
```

otaviojacobi (approved):

